PR #39, opened 2020-10-22
FIX #3 issue : Add much more key in KeyCode struct

### LINKED ISSUE #3: Add much more key in KeyCode struct
these keyCodes's value are from SDL_scancode.h
check also: https://wiki.libsdl.org/SDLScancodeLookup

aristid-lavri:
This fix #3 issue : Add much more key in KeyCode struct


JeanPhilippeKernel:
Great !! 🎉🎉 thanks a lot !

PR #215, opened 2023-06-19
Vulkan pipeline integration + Resources management revision

### LINKED ISSUE #216: Allocated buffers (Device memory, Image, ImageView, etc) don't get release at exit time gracefully
Repro : 
In Debug mode
launch engine -> add entity -> add geometry component ->  close the engine

### LINKED ISSUE #218: Remove Glad dependency

JeanPhilippeKernel:
Initially, this PR intended to introduce the support of the first Vulkan pipeline into ZEngine.
but, during the review process, it turned out that I should extend it to engine resource management.
This explains why this PR is now **big**. (reviewers should pardon me 😅😅)

Find below the areas that have been impacted during this incremental revision: 

## Shaders
Before, shader codes (vertex and fragment) were in the same file and compiled on runtime, slowing the engine startup.
I decided to split it into 2 separate files (.vert and .frag) and compile them during the engine compilation process.

## Core Engine and Renderer resources management
Some resources have been revisited for better memory management and ease of consumption of the Vulkan APIs.
- Separation Window and Swapchain implementation
- Made all Vulkan Device API static, and merged into one step the creation and initialization of Instance and logical device
- Revisited and fixed bugs around Buffer-type objects (Vertex, Uniform, Index, Framebuffer, Command Buffer)
- Introduced abstraction of CommandPool, Image, Texture, Attachment, (aka VkRenderPass) Pipeline, Fence, Semaphore and their associate specifications
- Introduced RenderPass that describes the Passes process used by the Renderers

## Other fixes and cleanup
- Fixed the CMake process waiting indefinitely after the build
- Removed GLAD dependency from externals.cmake definition
- Updated clang-format file 
- cleaned up dead codes /files


ZEngine/src/VulkanInstance.cpp
```diff
@@ -54,7 +54,8 @@ namespace ZEngine::Hardwares
         // std::copy(std::begin(validation_layer_name_collection), std::end(validation_layer_name_collection), std::back_inserter(instance_layer_name_collection));
 #endif
 
-        std::unordered_set<std::string> extension_layer_name_collection = {VK_KHR_SURFACE_EXTENSION_NAME, VK_KHR_WIN32_SURFACE_EXTENSION_NAME, VK_EXT_DEBUG_UTILS_EXTENSION_NAME};
+        std::unordered_set<std::string> extension_layer_name_collection = {
+            VK_KHR_SURFACE_EXTENSION_NAME, VK_KHR_WIN32_SURFACE_EXTENSION_NAME, VK_EXT_DEBUG_UTILS_EXTENSION_NAME};
 
         auto layer_properties = m_layer.GetInstanceLayerProperties();
         for (const LayerProperty& layer : layer_properties)
```
Sykzes: This is not how you should choose layers to enable. Depending on the machine, enabling all available layers and associated extensions might not even work since they may not be compatible. As a practical example, with this code the instance creation fails on my pc due to an unsupported layer configuration.

What you rather should do is to decide beforehand which layers you need for your application, such as VK_LAYER_LUNARG_api_dump that you specify above, verify if the instance actually supports the layer using GetInstanceLayerProperties, and finally create the instance with only the layers you need and are supported.

ZEngine/src/VulkanDevice.cpp
```diff
@@ -110,35 +92,63 @@ namespace ZEngine::Hardwares
...
 
-        if (!m_graphic_queue_family_index_collection.empty())
-        {
-            vkGetDeviceQueue(m_logical_device, m_graphic_queue_family_index_collection[0], 0, &m_graphic_queue);
-        }
-
         if (!m_graphic_with_present_queue_family_index_collection.empty())
         {
-            vkGetDeviceQueue(m_logical_device, m_graphic_with_present_queue_family_index_collection[0], 0, &m_graphic_with_present_support_queue);
+            for (uint32_t family_index : m_graphic_with_present_queue_family_index_collection)
```
Sykzes: You can do this but I would consider that overkill. On my machine, this code creates 16 queues for family index 0 alone, which I doubt you will ever need for your rendering workloads. I'm not sure what you intent to do but generally, I rarely see more than one queue created of a specific type (graphics, transfer, compute).

The typical workflow to parallelize Vulkan CPU work is rather to have multiple threads record command buffers in parallel and then submit them to a single queue.

I've also seen that you specify all queue families when creating resources and enable VK_SHARING_MODE_CONCURRENT so every queue you created can access them simultaneously. Depening on the underlying Vulkan implementation and/or the hardware, this can result in lower performance than using a single queue and/or VK_SHARING_MODE_EXCLUSIVE.

I recommend you start with what you need for your current purposes, which is probably one graphics queue and maybe one transfer queue, if you want to offload staging and downloads from the graphics queue. But maybe just start with a graphics queue, you can do transfer there too.

ZEngine/src/VulkanDevice.cpp
```diff
@@ -152,63 +162,140 @@ namespace ZEngine::Hardwares
...
+    QueueView VulkanDevice::GetAnyGraphicQueue() const
     {
-        if (with_present_support)
-        {
-            assert(m_graphic_with_present_support_queue != VK_NULL_HANDLE);
-            return m_graphic_with_present_support_queue;
-        }
-        assert(m_graphic_queue != VK_NULL_HANDLE);
-        return m_graphic_queue;
+        ZENGINE_VALIDATE_ASSERT(!m_graphic_with_present_support_queue_collection.empty(), "Graphic Queue with Present support can't be empty")
```
Sykzes: It looks to me like your intent is to continously select a queue at random and submit different chunks of graphics work to them. This is not really how you should achieve parallelism in Vulkan. 

Like this, you're going to run into complicated synchronization issues very quickly, for example if you're having multiple random queues executing workloads of the same scene into the same framebuffers in parallel. Also, you're going to run into ordering issues of pixel writes. All synchronization has to be done manually in Vulkan, meaning the synchronization between queues as well. So if you're not expecting to have completely independent graphics workloads, multiple random queues will just make everything needlessly complicated.

ZEngine/src/ImguiLayer.cpp
```diff
@@ -63,10 +68,18 @@ namespace ZEngine::Layers
             ImGui_ImplGlfw_InitForVulkan(reinterpret_cast<GLFWwindow*>(current_window_ptr->GetNativeWindow()), false);
 
             // Create Descriptor Pool
-            std::vector<VkDescriptorPoolSize> pool_sizes = {{VK_DESCRIPTOR_TYPE_SAMPLER, 1000}, {VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER, 1000},
-                {VK_DESCRIPTOR_TYPE_SAMPLED_IMAGE, 1000}, {VK_DESCRIPTOR_TYPE_STORAGE_IMAGE, 1000}, {VK_DESCRIPTOR_TYPE_UNIFORM_TEXEL_BUFFER, 1000},
-                {VK_DESCRIPTOR_TYPE_STORAGE_TEXEL_BUFFER, 1000}, {VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER, 1000}, {VK_DESCRIPTOR_TYPE_STORAGE_BUFFER, 1000},
-                {VK_DESCRIPTOR_TYPE_UNIFORM_BUFFER_DYNAMIC, 1000}, {VK_DESCRIPTOR_TYPE_STORAGE_BUFFER_DYNAMIC, 1000}, {VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, 1000}};
+            std::vector<VkDescriptorPoolSize> pool_sizes = {
```
Sykzes: You can do it like this for the moment to get going with your implementation. Just be aware that creating a pool reserving n of all descriptor types might waste a lot of memory depending on the descriptor sets' layouts you allocate from that pool. 

For example, consider the case where you allocate 1000 descriptor sets with a layout containing only one sampled image. The descriptor pool will then not be able to allocate more descriptor sets of that layout or any layout containing sampled images. But the rest of the descriptor types are still reserved, so you have 10000 individual descriptors still reserved by that pool without using them.

ZEngine/src/Helper.cpp
```diff
@@ -0,0 +1,879 @@
...
+        {
+            ZENGINE_CORE_ERROR("Failed to find a valid memory type")
+        }
+
+        VkMemoryAllocateInfo memory_allocate_info = {};
+        memory_allocate_info.sType                = VK_STRUCTURE_TYPE_MEMORY_ALLOCATE_INFO;
+        memory_allocate_info.allocationSize       = memory_requirement.size;
+        memory_allocate_info.memoryTypeIndex      = memory_type_index;
+
+        // ToDo : Call of vkAllocateMemory(...) isn't optimal because it is limited by maxMemoryAllocationCount
```
Sykzes: Yes, what the comment says :)

The baseline maxMemoryAllocationCount the implementation has to guarantee to be Vulkan compliant is 1024 btw, so not that much all things considered.

ZEngine/src/Helper.cpp
```diff
@@ -0,0 +1,879 @@
...
+    {
+        ZENGINE_VALIDATE_ASSERT(vkEndCommandBuffer(command_buffer) == VK_SUCCESS, "Failed to end the Command Buffer")
+
+        VkSubmitInfo submit_info       = {};
+        submit_info.sType              = VK_STRUCTURE_TYPE_SUBMIT_INFO;
+        submit_info.commandBufferCount = 1;
+        submit_info.pCommandBuffers    = &command_buffer;
+
+        vkQueueSubmit(queue, 1, &submit_info, VK_NULL_HANDLE);
+        vkQueueWaitIdle(queue);
```
Sykzes: I think we talked about ...WaitIdle already, so just for completeness: These will block the CPU until the queue is emptied and also will reduce GPU occupancy, since the queues might never fill up with command buffers. It is basically lockstepping the CPU and GPU, so the better way would be to submit the work and check via synchronization objects like fences, timeline semaphores, events, etc. if the transfer has completed and the object is usable.

ZEngine/src/Helper.cpp
```diff
@@ -0,0 +1,879 @@
...
+        graphic_pipeline_create_info.pNext               = nullptr;        // Optional
+
+        ZENGINE_VALIDATE_ASSERT(
+            vkCreateGraphicsPipelines(device.GetNativeDeviceHandle(), VK_NULL_HANDLE, 1, &graphic_pipeline_create_info, nullptr, &out_graphic_pipeline) == VK_SUCCESS,
+            "Failed to create Graphics Pipeline")
+
+        // Cleanup ShaderModules
+        for (auto& shader_module : shader_module_collection)
+        {
+            vkDestroyShaderModule(device.GetNativeDeviceHandle(), shader_module, nullptr);
```
Sykzes: Not Vulkan specific but I would say it is a bit of an API contract problem if you pass in the pipline specification object as const& but then call destroy on the shader module handles contained in that specification. I also don't think you really need the ShaderModule member at all in the specification since the create infos already contain the handles.

ZEngine/src/Helper.cpp
```diff
@@ -0,0 +1,879 @@
...
+        specification.DepthStencilStateCreateInfo.minDepthBounds        = 0.0f; // Optional
+        specification.DepthStencilStateCreateInfo.maxDepthBounds        = 1.0f; // Optional
+        specification.DepthStencilStateCreateInfo.stencilTestEnable     = VK_FALSE;
+        specification.DepthStencilStateCreateInfo.front                 = {}; // Optional
+        specification.DepthStencilStateCreateInfo.back                  = {}; // Optiona
+
+        /*Color blending*/
+        // This configuration is per-framebuffer definition
+        specification.ColorBlendAttachmentState.colorWriteMask      = VK_COLOR_COMPONENT_R_BIT | VK_COLOR_COMPONENT_G_BIT | VK_COLOR_COMPONENT_B_BIT | VK_COLOR_COMPONENT_A_BIT;
+        specification.ColorBlendAttachmentState.blendEnable         = VK_TRUE;
```
Sykzes: I would consider blending disabled to be a better default state, since blending isn't that cheap and having it enabled always is something that can be easily missed and potentially only seen in a capture. For example, with the default blending factors here, there will be virtually no difference in output if your fragment shader always writes alpha == 1.0 with and without blending enabled, but the former will run slower.

ZEngine/src/Helper.cpp
```diff
@@ -0,0 +1,879 @@
...
+        specification.VertexInputStateCreateInfo.pVertexAttributeDescriptions    = nullptr;
+        specification.VertexInputStateCreateInfo.vertexBindingDescriptionCount   = 0;
+        specification.VertexInputStateCreateInfo.pVertexBindingDescriptions      = nullptr;
+        specification.VertexInputStateCreateInfo.flags                           = 0;
+        specification.VertexInputStateCreateInfo.pNext                           = nullptr;
+    }
+
+    void TransitionImageLayout(Hardwares::VulkanDevice& device, VkImage image, VkFormat image_format, VkImageLayout old_image_layout, VkImageLayout new_image_layout)
+    {
+        auto                 graphic_queue  = device.GetAnyGraphicQueue();
```
Sykzes: There are multiple issues I'm seeing with TransitionImageLayout:

1. You shouldn't submit command buffers containing only single transitions. Rather, layout transitions are supposed to occur between different sets of commands in the same command buffer, or different command buffers on the same queue, or different command buffers on different queues. While it is not wrong to use them like this it is also wasteful to submit a single command buffer for each transition. Try including the transitions in your other command buffers, such as the transfer command buffer.
2. The way I understand GetAnyGraphicQueue is that each time a different queue can be returned. The only reason this works right now is because you are lockstepping the submits. Imagine the function being used to transition the image layout multiple times, but not waiting for the queues to be idle. Which transition would be executed first? In which layout would the image then be?

ZEngine/include/ZEngine/Rendering/Buffers/IndexBuffer.h
```diff
@@ -41,18 +56,131 @@ namespace ZEngine::Rendering::Buffers
...
+            staging_buffer,
+            staging_buffer_memory,
+            static_cast<VkDeviceSize>(this->m_byte_size),
+            VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
+            memory_properties,
+            VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT);
+
+        void* memory_region;
+        ZENGINE_VALIDATE_ASSERT(vkMapMemory(device_handle, staging_buffer_memory, 0, this->m_byte_size, 0, &memory_region) == VK_SUCCESS, "Failed to map the memory")
+        std::memcpy(memory_region, content.data(), this->m_byte_size);
```
Sykzes: I ran into a crash here. It looks like content might be empty sometimes, so content.data() is nullptr. Same goes for the other variants probably.

Sykzes: I think the VertexBuffer class exhibits the same issue.

ZEngine/src/GlfwWindow/VulkanWindow.cpp
```diff
@@ -420,10 +439,89 @@ namespace ZEngine::Window::GLFWWindow
...
+        VkResult acquire_image_result = vkAcquireNextImageKHR(device_handle, m_swapchain, UINT64_MAX, current_frame.ImageAcquiredSemaphore, VK_NULL_HANDLE, &image_index);
+        if (acquire_image_result == VK_ERROR_OUT_OF_DATE_KHR || acquire_image_result == VK_SUBOPTIMAL_KHR || acquire_image_result == VK_NOT_READY || acquire_image_result == VK_TIMEOUT)
+        {
+            return;
+        }
+        ZENGINE_VALIDATE_ASSERT(acquire_image_result == VK_SUCCESS, "Failed to acquire image from Swapchain")
+
+        if (!current_frame.GraphicCommandBuffers.empty())
+        {
+            vkFreeCommandBuffers(
```
Sykzes: Technically, you could ommit this, since you also call vkResetCommandPool on the whole pool, which resets all allocated command buffers from that pool back to the initial state. So you can immediately call vkBeginCommandBuffer on any of them again to record new commands without freeing and reallocating them first.

For completeness: Another option would be to not use vkResetCommandPool, but reset the individual command buffers using vkResetCommandBuffer and record into them again with vkBeginCommandBuffer. You still won't need to free them this way. Caveat: this requires the VK_COMMAND_POOL_CREATE_RESET_COMMAND_BUFFER_BIT flag during the CommandPool's creation.

ZEngine/include/ZEngine/Rendering/Buffers/IndexBuffer.h
JeanPhilippeKernel: Interesting. 
can you help with a repro case?

ZEngine/src/VulkanInstance.cpp
JeanPhilippeKernel: Thanks !!
I updated the implementation,
lemme know if it fixes the crash on your side

Sykzes: Still fails instance creation. Did you push your changes? I pulled from vulkan-pipeline but to me it looks like the same code is still in there, meaning you still enable all instance layers that get returned by GetInstanceLayerProperties.

ZEngine/src/VulkanInstance.cpp
```diff
@@ -125,27 +130,36 @@ namespace ZEngine::Hardwares
 
         for (const VkPhysicalDevice& physical_device : physical_device_collection)
         {
+            std::unordered_set<std::string_view> device_enabled_layer_name_collection;
+            std::unordered_set<std::string_view> device_extension_layer_name_collection = {VK_KHR_SWAPCHAIN_EXTENSION_NAME};
+
             for (LayerProperty& layer : layer_properties)
```
Sykzes: This also doesn't work on my machine. Same as for the instance creation, enabling all layers and extensions available on the physical device is not a good idea. Decide in advance, which layers and extensions you need (if any), check if they are available here and then only enable these, or error out.

ZEngine/src/VulkanInstance.cpp
JeanPhilippeKernel: Pushed another revision of the impl., by selecting the required layers, hope that fixes the crash from your side 🤞

Sykzes: It now fails to compile. Looks like m_selected_layer_property_collection is missing in the class's header. I think you forgot to push VulkanInstance.h.

In terms of the actual issue, I added m_selected_layer_property_collection locally to get it to compile and instance creation now works, so beside the compiler error, looks good to me now.

JeanPhilippeKernel: Doh !🤦my bad.. pushed now ... thanks !!!

ZEngine/src/VulkanDevice.cpp
JeanPhilippeKernel: thanks a lot for the feedback.
I revisited the whole engine resource management to stick with only needed queue (transfer, graphic)  and maybe later compute

ZEngine/src/Helper.cpp
JeanPhilippeKernel: Thanks a lot !
I removed the use of vkQueueWaitIdle and switch to Fence and Semaphore

ZEngine/src/GlfwWindow/VulkanWindow.cpp
JeanPhilippeKernel: Thanks a lot !,
I think this is no longer applicable, given the fact I revisited the whole Vulkan resource management.
we might need to review it again  -- sorry for this huge PR 😅

ZEngine/src/Helper.cpp
JeanPhilippeKernel: Removed those apis.
thanks!

ZEngine/src/VulkanDevice.cpp
JeanPhilippeKernel: thanks a lot for the feedback.
I revisited the whole engine resource management to stick with only the needed queue (transfer, graphic) and maybe later compute

ZEngine/src/VulkanInstance.cpp
JeanPhilippeKernel: Completed removed this file and its implementation.
Thanks a lot !

ZEngine/src/Helper.cpp
JeanPhilippeKernel: Fixed, thanks a lot !

ZEngine/src/Helper.cpp
JeanPhilippeKernel: Fixed :) ... Thanks

ZEngine/include/ZEngine/Rendering/Buffers/IndexBuffer.h
JeanPhilippeKernel: Fixed :) ... thanks

ZEngine/src/Helper.cpp
JeanPhilippeKernel: I created a issue to track it: https://github.com/JeanPhilippeKernel/RendererEngine/issues/221

ZEngine/src/ImguiLayer.cpp
JeanPhilippeKernel: Thanks a lot! , Fixed now
This was coming from the ImGui example.

ZEngine/src/CommandPool.cpp
```diff
@@ -0,0 +1,87 @@
...
+            m_current_command_buffer_index = 0;
+
+            auto device = Hardwares::VulkanDevice::GetNativeDeviceHandle();
+            for (auto command : m_command_buffer_collection)
+            {
+                ZENGINE_VALIDATE_ASSERT(command->GetState() != Buffers::CommanBufferState::Recording, "")
+
+                if (command->GetState() == Buffers::CommanBufferState::Submitted)
+                {
+                    command->WaitForExecution();
```
Sykzes, line 62: For visibility as discussed: This can create situations where the cpu unnecessarily waits instead of doing work (when the 10 command buffers are all used). Also, consider the case where a commandbuffer has finished execution and could be reused.

For the future, consider waiting for at most one to get going again, or more advanced strategies which do not wait for a command buffer execution at all but uses new command buffers and/or command pools when all existing are executing.

On a side note, you could run into problems when WaitForExecution times out, as you are calling vkResetCommandPool down below, which kills potentially executing command buffers. In the worst case, this can cause a lost device.

ZEngine/src/CommandBuffer.cpp
```diff
@@ -0,0 +1,106 @@
...
+#include <Hardwares/VulkanDevice.h>
+
+namespace ZEngine::Rendering::Buffers
+{
+    CommandBuffer::CommandBuffer(VkCommandPool command_pool, Rendering::QueueType type, bool present_on_swapchain)
+    {
+        m_queue_type = type;
+        if (present_on_swapchain)
+        {
+            m_signal_semaphore = CreateRef<Primitives::Semaphore>();
```
Sykzes, line 13: As discussed, for visibility: you technically only need a semaphore per submit, not per command buffer. 

Keep in mind, you need a semaphore if you want a coarse-granular wait of one action on a queue to another action on a queue. So you could also have semaphores to synchronize submit-to-submit, not only submits-to-present. So the present_on_swapchain guard might not make sense in the future.

ZEngine/src/Attachment.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+namespace ZEngine::Rendering::Renderers::RenderPasses
+{
+    Attachment::Attachment(const Specifications::AttachmentSpecification& spec)
+    {
+        ZENGINE_VALIDATE_ASSERT(!spec.ColorsMap.empty(), "Color attachments can't be empty")
+
+        VkSubpassDescription                 subpass_description                   = {};
+        std::vector<VkAttachmentDescription> attachment_description_collection     = {};
+        VkAttachmentReference                depth_attachment_reference            = {.attachment = UINT32_MAX, .layout = VK_IMAGE_LAYOUT_UNDEFINED};
```
Sykzes, line 15: Could use VK_ATTACHMENT_UNUSED, which is the same as UINT32_MAX incidentially, but more in line with the API.

Sykzes (approved):

ZEngine/src/Attachment.cpp
JeanPhilippeKernel: Thanks @Sykzes for your valuable evaluation and feedback since Day One :)

ZEngine/src/CommandBuffer.cpp
JeanPhilippeKernel: Thanks a lot for those comments @Sykzes :)

Having those ones shaped the design and usage CommandBuffer of the engine in 2026 !!! -

ZEngine/src/CommandPool.cpp
JeanPhilippeKernel: Same for this comment :) !! 💯

PR #317, opened 2024-04-06
Runtime Shader compiler API

### LINKED ISSUE #269: Improve Shader Compiler
Today, GLSL shader codes are compiled offline during the engine build stage through  glslangValidator tool.

In order to enhance and improve the writing shader experience, we would like to support shader compilation on runtime.
That support will be the foundation of other features like:
+ Writing shader in-engine editor
+ Hot-reloading compiled shader

The offline compilation leverages the support _**GL_GOOGLE_include_directive**_ extension that allows the inclusion of another shader file, the runtime compiler should offer/allow the same feature to preserve the same experience.
 
The existing Shader Compiler was based on OpenGL API, but since we moved to Vulkan we need to rethink the exposed APIs 

- [x] Revisit the existing Shader Compiler to make it compatible with the Vulkan shader compiler API
- [x] Add support for #include<....>
- [x] Add support for runtime compilation

jnyfah:
This PR introduces a runtime shader compiler API, enabling on-the-fly GLSL shader compilation. 

Key features include:

- GL_GOOGLE_include_directive Support: The API fully supports the `#include` directive within GLSL shaders. It is assumed that all included files reside in the same directories as their parent shaders.

- Shader Validation and Optimization: Post-compilation, the API validates shaders to ensure they meet the SPIR-V specification and applies optimizations focused on reducing shader size

- Resource Directory Convention: The system expects all shader source files to be located within a designated resource directory (`Resources/Shaders`)

- Includes tests to verify shader compilation, without extending to functional correctness of the shaders


ZEngine/src/CMakeLists.txt
```diff
@@ -49,4 +49,5 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
 	target_link_libraries(zEngineLib PRIVATE stdc++fs)
 endif ()
 
-target_link_libraries (zEngineLib PUBLIC imported::External_libs)
\ No newline at end of file
+target_link_libraries (zEngineLib PUBLIC imported::External_libs)
+target_link_libraries (zEngineLib PRIVATE glslang SPIRV SPIRV-Tools)
```
JeanPhilippeKernel: you don't need this link here since you already done in externals.cmake

ZEngine/include/ZEngine/Rendering/Shaders/Compilers/LinkageStage.h
```diff
@@ -18,13 +19,16 @@ namespace ZEngine::Rendering::Shaders::Compilers
...
         /**
          * Run asynchronously compiler stage
          *
          * @param information Collection of shader information
          */
-        virtual std::future<void> RunAsync(std::vector<ShaderInformation>& information) override;
+        virtual std::future<void> RunAsync(ShaderInformation& information) override;
+
+    private:
+        const char* outputDirectory = "Shaders/Cache/";
```
JeanPhilippeKernel: use string_view ?

ZEngine/include/ZEngine/Rendering/Shaders/ShaderIncluder.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace ZEngine::Rendering::Shaders
+{
+
+    class ShaderIncluder : public glslang::TShader::Includer
+    {
+    public:
+        ShaderIncluder() {}
+
+        IncludeResult* includeLocal(const char* headerName, const char* includerName, size_t inclusionDepth) override
```
JeanPhilippeKernel, line 17: use of string_view instead of const char * ?

ZEngine/include/ZEngine/Rendering/Shaders/ShaderIncluder.h
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        void releaseInclude(IncludeResult* result) override
+        {
+            delete result;
+        }
+
+    private:
+        const char*                                                   basePath = "Shaders/";
+        std::unordered_map<std::string, std::shared_ptr<std::string>> cache;
```
JeanPhilippeKernel: not sure I understand the use of shared_ptr<std::string> here.

ZEngine/src/ShaderReader.cpp
```diff
@@ -40,11 +40,21 @@ namespace ZEngine::Rendering::Shaders
         return buffer;
     }
 
+    ShaderType ShaderReader::GetShaderType(const std::filesystem::path& path)
+    {
+        if (path.extension() == ".vert")
+            return ShaderType::VERTEX;
+        if (path.extension() == ".frag")
+            return ShaderType::FRAGMENT;
+        if (path.extension() == ".frag")
```
JeanPhilippeKernel: you mean ".geom" for geometry shader

ZEngine/include/ZEngine/Rendering/Shaders/Compilers/CompilationStage.h
```diff
@@ -13,18 +16,44 @@ namespace ZEngine::Rendering::Shaders::Compilers
...
-        virtual std::future<void> RunAsync(std::vector<ShaderInformation>& information) override;
+        virtual std::future<void> RunAsync(ShaderInformation& information) override;
+
+        EShLanguage GetEShLanguage(const ShaderType type);
+
+        void SetShaderRules(glslang::TShader& shader, ShaderInformation& information_list, TBuiltInResource& Resources);
+
     };
+
+    class GlslangInitializer
```
JeanPhilippeKernel, line 32: not sure I see where you use this class?

ZEngine/src/CompilationStage.cpp
```diff
@@ -1,25 +1,94 @@
...
+
+        ShaderIncluder includefiles;
+
+        if (!shader.parse(&Resources, 100, false, messages, includefiles))
+        {
+            m_information = {false, shader.getInfoLog()};
+            co_return;
+        }
+
+        glslang::TProgram program;
```
JeanPhilippeKernel, line 83: it seems to that this part is the linkage step... right ?  if yes I think we should move it to the linkage stage impl

ZEngine/src/ShaderCompiler.cpp
```diff
@@ -32,57 +32,44 @@ namespace ZEngine::Rendering::Shaders::Compilers
...
-        bool compile_process_succeeded{true};
-
-        std::vector<ShaderInformation> shader_information;
-
         ShaderOperationResult read_operation = co_await m_reader->ReadAsync(m_source_file);
         if (read_operation == ShaderOperationResult::FAILURE)
         {
             ZENGINE_CORE_CRITICAL("Compilation process stopped")
-            co_return std::make_tuple(ShaderOperationResult::FAILURE, std::vector<ShaderInformation>{});
+            co_return std::make_tuple(ShaderOperationResult::FAILURE, ShaderInformation{});
```
JeanPhilippeKernel: I'd be happy if we can get rid of tuple usage for the performance's sake. 
We can replace it by our own defined struct.

JeanPhilippeKernel:
I think we don't need the new introduced external deps you added in this PR - the existing ones already contain the headers you need for the work

ZEngine/include/ZEngine/Rendering/Shaders/ShaderIncluder.h
jnyfah: `includeLocal` function is an override of a virtual function defined in `glslang::TShader::Includer`, so the function must match exactly what is declared in the base class

ZEngine/src/CompilationStage.cpp
jnyfah: nope its not

ZEngine/include/ZEngine/Rendering/Shaders/Compilers/CompilationStage.h
jnyfah: this is just a RAII implementation of glslang, just like smart pointers, to ensure its finalized every time it is initialized

used here: [CompilationStage.cpp](https://github.com/jnyfah/RendererEngine/blob/b6e7aee04f3897f04e73afd670d5d7eb5298d0ce/ZEngine/src/CompilationStage.cpp#L59-L64)

` GlslangInitializer glslangInit;`
`if (!glslangInit.isSuccess())
   {
        m_information = {false, "Failed to initialize glslang"};
            co_return;
}
`

JeanPhilippeKernel (approved):

PR #326, opened 2024-05-30
Nuget script to automate the installation of nuget 

### LINKED ISSUE #323: Automate the installation of NuGet on dev machine

jnyfah:
closes out #323 , on the first build user has to grant admin privileges when prompted


Scripts/Nuget.ps1
JeanPhilippeKernel, line 1: I think we should split the whole implementation of that NuGet as follow: 
(1) a function to setup the NuGet, something like Setup-NuGet 
(2) a function to find the NuGet, something like Find-NuGet

All of them should be in Shared.ps1 like other funcs

Scripts/Nuget.ps1
```diff
@@ -0,0 +1,47 @@
...
+    exit
+}
+
+# Create the directory if it doesn't exist
+if (-not (Test-Path $installPath)) {
+    New-Item -ItemType Directory -Path $installPath
+}
+
+# Download nuget.exe since it does not exist
+Invoke-WebRequest -Uri "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe" -OutFile $nugetPath
```
JeanPhilippeKernel: Does the URL never change?  if no, we may want to update the repoConfiguration.json 
Do we have win-x64? if yes, let use the x64 version

Scripts/Nuget.ps1
```diff
@@ -0,0 +1,47 @@
...
+
+# Function to add a path to the system PATH environment variable 
+$scriptBlock = {
+    function Add-ToSystemPath {
+        $installPath = "$env:USERPROFILE\NuGet"
```
JeanPhilippeKernel: Use Join-Path instead of "\\"

Scripts/Nuget.ps1
```diff
@@ -0,0 +1,47 @@
...
+        else {
+            Write-Host "Path already exists in system PATH: $installPath"
+        }
+    }
+
+    Add-ToSystemPath
+}
+
+# Add NuGet to system PATH
+$encodedScript = [Convert]::ToBase64String([Text.Encoding]::Unicode.GetBytes($scriptBlock.ToString()))
```
JeanPhilippeKernel: Not sure I understand the encoding process

Scripts/BuildEngine.ps1
```diff
@@ -55,6 +55,8 @@ $RepoRoot = [IO.Path]::Combine($PSScriptRoot, "..")
 Write-Host "Ensuring submodules are initialized and updated..."
 git -C $RepoRoot submodule update --init --recursive
 
+. (Join-Path $PSScriptRoot Nuget.ps1)
```
JeanPhilippeKernel: we might no longer need this if we move the whole impl. to Shared.ps1

JeanPhilippeKernel (changes requested):

JeanPhilippeKernel (approved):

PR #328, opened 2024-06-06
Improved Nuget program detection

### LINKED ISSUE #327: Installing/Detecting NuGet program can be improved
Currently, when searching for the NuGet program, it targets a specific directory and if not present a perform the download. However, there may be situations where developers have installed it elsewhere and made it accessible from the terminal. 

For this task, we aim to locate the NuGet program, having defined a candidate for it.
See CMake example: https://github.com/JeanPhilippeKernel/RendererEngine/blob/develop/Scripts/Shared.ps1#L99

jnyfah:
Closes #327

### DIFF
--- a/Scripts/Shared.ps1
+++ b/Scripts/Shared.ps1
@@ -42,8 +42,11 @@ function Find-Nuget () {
     $NugetMinimumVersion = $repoConfiguration.Requirements.Nuget.Version
 
     $NugetProgramCandidates = @(
-        'nuget'
-        if ($IsWindows) { 'nuget.exe' }
+        'nuget' 
+        Join-Path -Path $repoConfiguration.Paths.Tools -ChildPath 'NuGet\nuget'
+        if ($IsWindows) {
+          Join-Path -Path $repoConfiguration.Paths.Tools -ChildPath 'NuGet\nuget.exe'
+        }
     )
 
     foreach ($NugetProgram in $NugetProgramCandidates) {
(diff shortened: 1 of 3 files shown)


Scripts/Shared.ps1
```diff
@@ -35,19 +35,42 @@ function Get-RepositoryToolPath () {
...
+function Find-Nuget () {
 
-    return $true
+    $repoConfiguration = Get-RepositoryConfiguration
+    $NugetMinimumVersion = $repoConfiguration.Requirements.Nuget.Version
+
+    $NugetProgramCandidates = @(
+        'nuget'
+        if ($IsWindows) {
+            Join-Path -Path $env:LOCALAPPDATA -ChildPath 'nuget\nuget.exe'
```
JeanPhilippeKernel: I think we should simplify this whole part for Windows plat., since we can't guess the right path, just having 'nuget.exe' as option if enough.

`if($IsWindows) {
    'nuget.exe'
}`

JeanPhilippeKernel (changes requested):

Scripts/Shared.ps1
JeanPhilippeKernel: Okay it's failing because after our download and install we dont point to our NuGet.

let's make thing simple and scope it to the project only:
(1) For the detection
let's look for 
```pwsh
@(
    'nuget' 
    Join-Path -Path $repoConfiguration.Paths.Tools -ChildPath 'NuGet\nuget'
    if ($IsWindows) {
      'nuget.exe'
      Join-Path -Path $repoConfiguration.Paths.Tools -ChildPath 'NuGet\nuget.exe'
    }
)
`

File: Scripts/Shared.ps1 [hunk 2]
```diff
@@ -35,36 +35,48 @@ function Get-RepositoryToolPath () {
...
 }
 
 function Setup-NuGet {
-    param([string]$nugetPath, [string]$installPath)
-    try {
-        if (-not (Test-Path $installPath)) { 
-            New-Item -ItemType Directory -Path $installPath | Out-Null
-        }
+    $installPath = Join-Path -Path (Get-RepositoryToolPath) -ChildPath "NuGet"
+    $nugetPath = Join-Path -Path $installPath -ChildPath "nuget.exe"
```
JeanPhilippeKernel:
```suggestion
    $nugetProgramName = If($IsWindows) {"nuget.exe"} Else {"nuget"}
    $nugetPath = Join-Path -Path $installPath -ChildPath $nugetProgramName
```

JeanPhilippeKernel (changes requested):

JeanPhilippeKernel (approved):

PR #335, opened 2024-06-20
fix path error

BRICEMARCELKOUADIO:
-Impossible to build the project when the language is French because line 55 of Shared.ps1 "(& $Command help | Select-String -Pattern "NuGet Version" | Out-String)" has the value NuGet Version :... in French and NuGet Version:... in English, we see that there is a space before the 2 points of the NuGet Version in French. so the condition will always return false for people who have their configuration in French. With -ForceEnglishOutput, we ensure that the output is in English.

### DIFF
--- a/Scripts/BuildEngine.ps1
+++ b/Scripts/BuildEngine.ps1
@@ -68,9 +68,11 @@ else {
     }
 }
 
-# Add NuGet to the PATH for the current session
+# Add NuGet to the PATH for the current session if it's not already there
 $installPath = Split-Path -Path $nugetProgram -Parent
-$env:PATH = "$installPath;$env:PATH"
+if ($env:PATH -notlike "*$installPath*") {
+    $env:PATH = "$installPath;$env:PATH"
+}
 
 $RepoRoot = [IO.Path]::Combine($PSScriptRoot, "..")
 Write-Host "Ensuring submodules are initialized and updated..."
--- a/Scripts/BuildEngine.ps1
+++ b/Scripts/BuildEngine.ps1
@@ -68,6 +68,10 @@ else {
     }
 }
 
+# Add NuGet to the PATH for the current session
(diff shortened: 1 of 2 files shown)


Scripts/Shared.ps1
```diff
@@ -75,9 +75,26 @@ function Setup-NuGet {
...
+    if ($path -notlike "*$installPath*") {
+        $newPath = "$path;$installPath"
+        [Environment]::SetEnvironmentVariable("Path", $newPath, [EnvironmentVariableTarget]::User)
+        # Update the current session's PATH environment variable
+        $env:Path += ";$installPath"
+        Write-Host "Path added to user PATH: $installPath"
+    } else {
+        Write-Host "Path already exists in user PATH: $installPath"
+    }
+}
```
jnyfah: I dont think we need this as adding the NuGet path to the system environment paths with each new build causes an accumulation of outdated paths, especially after clean builds.

so to avoid this, we should modify our approach to add the NuGet path only to the current PowerShell process environment. This would ensure the path is available during the current session but does not persist, preventing the build-up of obsolete paths.

i would push it to this PR, apart from that, everything looks good to me 🙂

Scripts/BuildEngine.ps1
```diff
@@ -68,6 +68,10 @@ else {
     }
 }
 
+# Add NuGet to the PATH for the current session
+$installPath = Split-Path -Path $nugetProgram -Parent
+$env:PATH = "$installPath;$env:PATH"
```
JeanPhilippeKernel: is doing this won't continually concatenate the Nuget Path every run? let's say I run the script 3 times

jnyfah: okay, we can check if its already there before adding it to path

JeanPhilippeKernel (approved):

jnyfah (approved):

PR #361, opened 2024-07-26
Adding new build option to compile Panzerfaust only

### LINKED ISSUE #358: Separate Panzefaust and ZEngine build
As today, to build Panzerfaust, developers have to run the build engine script which will go through the whole build process, that depending on the platform can be extremely long.

The idea here is to introduce, ideally a switch option to the current build script that will skip the engine build and and only build panzerfaust.

This can look like: BuildEngine.ps1 -OnlyLauncher

aristid-lavri:
- Adding $LauncherOnly parameter to BuildEngine.ps1 that avoid building renderer engine when dev want to work on launcher only
- Adding new cmake option in cmakelist to that is used to avoid building renderer engine when $LauncherOnly is set as parameter in build command.
- Add $LauncherOnly related to build command parameter to avoid Postbuild operation

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,39 +18,42 @@ include(${ENLISTMENT_ROOT}/Scripts/CMake/NuGet.cmake)
 
 
 option (COPY_EXAMPLE_PROJECT "Copy example projects that show how to use Launcher" ON)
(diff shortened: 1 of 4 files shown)


jnyfah:
I think we should also update the readme

jnyfah (approved):

PR #364, opened 2024-08-07
Refactoring tests definition

### LINKED ISSUE #348: Failing to start unit test in pipeline
ZEngine tests at Run Test stage are failing to run in pipeline as it appears that it somehow has dependency on Vulkan SDK, which we no longer install 

See when we remove the Vulkan SDK from the pipeline: https://github.com/JeanPhilippeKernel/RendererEngine/actions/runs/9897311532/job/27341489380

See when we add the Vulkan SDK to the pipeline:
https://github.com/JeanPhilippeKernel/RendererEngine/actions/runs/9897816753/job/27343198356

JeanPhilippeKernel:
During the investigation,
I found that debug version of the test was failing to run due to missing vulkan-1.dll
The program at load time is looking for that .dll  based on env. path but it seems the container we run on fail to provide it.

Since we build and produce vulkan-1.dll at compile time, we updated the env path with the output lib path.


jnyfah:
Finally !! 🚀

jnyfah (approved):

PR #384, opened 2024-10-29
Reshaping IntrusivePtr public APIs

JeanPhilippeKernel:
In this PR we are reshaping the intrusive Ptr public api.
we moved the defined namespace aliases (Ref, Scope) into Helpers/IntrusivePtr header, and this has the consequences of updating all the codebase to consume the new namespace.

### DIFF
--- a/Tetragrama/Components/DockspaceUIComponent.cpp
+++ b/Tetragrama/Components/DockspaceUIComponent.cpp
@@ -12,6 +12,8 @@
 
 namespace fs = std::filesystem;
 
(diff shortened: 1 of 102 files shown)


jnyfah (approved):

jnyfah:
@JeanPhilippeKernel  why is the macOS build failing

JeanPhilippeKernel:
Fixed the issue, missed calling explicitly the namespace in IDevice helper.  Typical clang compiler behavior: it likes to complain about everything 😅

PR #426, opened 2025-03-10
Introducing Memory Manager

JeanPhilippeKernel:
This PR introduces Memory Manager and Pool Allocator.
In addition, we improved the implementation of Arena


jnyfah (approved):

jnyfah:
@JeanPhilippeKernel  why do we have allocator.cpp in the test cmake ?

PR #432, opened 2025-03-25
Low Budget Hash Map

jnyfah:
This will need to be revisited in the nearest future as dynamic growth means rehashing the entire hashmap again (hopefully for now that doesn't happen too often as the size doubles and the user also specifies capacity upfront)

Also removed size initialization from array init 

Added initializer list

### DIFF
--- a/ZEngine/ZEngine/Core/Containers/HashMap.h
+++ b/ZEngine/ZEngine/Core/Containers/HashMap.h
@@ -83,7 +83,7 @@ namespace ZEngine::Core::Containers
             if (m_size >= m_capacity * 0.75)
             {
                 reserve(m_capacity ? m_capacity * 2 : 8);
-                index = hash(key); // Recompute after reserve
+                index = hash(key);
             }
 
             Node* new_node     = static_cast<Node*>(ZAlloc(m_allocator, sizeof(Node), ZAlignof(Node)));
@@ -147,13 +147,7 @@ namespace ZEngine::Core::Containers
         {
             for (size_t i = 0; i < m_capacity; ++i)
             {
-                Node* current = m_data[i].head;
-                while (current)
-                {
-                    Node* next = current->next;
-                    // Optionally free current node here
-                    current    = next;
-                }
+                Node* current  = m_data[i].head;
                 m_data[i].head = nullptr;
             }
             m_size = 0;
@@ -176,7 +170,7 @@ namespace ZEngine::Core::Containers
                 {
                     Node*  next            = current->next;
 
-                    size_t new_index       = hash(current->key); // Now uses new m_capacity
+                    size_t new_index       = hash(current->key);
                     current->next          = m_data[new_index].head;
                     m_data[new_index].head = current;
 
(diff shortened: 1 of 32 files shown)


ZEngine/tests/initializerlist_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+
+TEST_F(InitializerListTest, WithArray)
+{
+    Array<int> array;
+    array.init(&arena, 4);
+    array.push(1);
+    array.push(2);
+    array.push(3);
+    array.push(4);
+
```
JeanPhilippeKernel: I'd expect the `initializer_list` to be used for this particular case:
e.g
`cpp
Array<int> array;
array.init(&arena, 4,  {1, 2, 3, 4})
`

ZEngine/tests/initializerlist_test.cpp
```diff
@@ -0,0 +1,84 @@
...
+    EXPECT_EQ(total, 10);
+}
+
+TEST_F(InitializerListTest, WithStrings)
+{
+    Array<String> str_array;
+    str_array.init(&arena, 3);
+    str_array.push(String());
+    str_array.push(String());
+    str_array.push(String());
```
JeanPhilippeKernel: Same comment about the expectation above

ZEngine/tests/hashmap_test.cpp
```diff
@@ -0,0 +1,157 @@
...
+    map.insert(5, 50);
+
+    EXPECT_TRUE(map.contains(1));
+    EXPECT_TRUE(map.contains(3));
+    EXPECT_TRUE(map.contains(5));
+
+    EXPECT_EQ(map[1], 10);
+    EXPECT_EQ(map[3], 30);
+    EXPECT_EQ(map[5], 50);
+}
```
JeanPhilippeKernel, line 263: Can you add tests with a user-defined struct/class for the hasmap ?

JeanPhilippeKernel (changes requested):

ZEngine/ZEngine/Core/Containers/HashMap.h
```diff
@@ -0,0 +1,205 @@
+#pragma once
+#include <Allocator.h>
+#include <MemoryOperations.h>
+#include <rapidhash.h>
+
+using namespace ZEngine::Core::Memory;
+
+namespace ZEngine::Core::Containers
+{
+    template <typename K, typename V>
```
JeanPhilippeKernel, line 10: Can we have HashMapView ? that will for iteration ?

JeanPhilippeKernel (changes requested):

ZEngine/tests/initializerlist_test.cpp
jnyfah: The issue is that `{1, 2, 3, 4}` syntax is special in C++ - it's directly tied to `std::initializer_list` but i can have a `make_initializer_list` variadic function as a helper

JeanPhilippeKernel (approved):

PR #447, opened 2025-04-25
feat(panzerfaust): add search support allowing user to find projet by name (#399)

### LINKED ISSUE #399: Implement project search feature
Provide users the possibility to search their projects by name

AmanMenda:
I've tried to keep the code as simple as possible while using the recommended types. Some comments left in the code to justify some of my choices :)

### DIFF
--- a/Panzerfaust/ViewModels/MainWindowViewModel.cs
+++ b/Panzerfaust/ViewModels/MainWindowViewModel.cs
@@ -1,5 +1,6 @@
 ﻿using DynamicData;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
 using Panzerfaust.Models;
 using ReactiveUI;
 using System;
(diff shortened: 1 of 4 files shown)


Panzerfaust/ViewModels/MainWindowViewModel.cs
```diff
@@ -18,27 +19,63 @@ namespace Panzerfaust.ViewModels
...
             CreateProjectCommand = ReactiveCommand.CreateFromTask(OnCreateProjectCommand);
 
             MessageBus.Current.Listen<(string, ProjectViewModel)>().Subscribe(OnReceiveMessage);
         }
 
+        private static Func<ProjectViewModel, bool> CreateFilterPredicate(string searchTerm)
+        {
+            return string.IsNullOrWhiteSpace(searchTerm)
+                ? _ => true
+                : p => p.Name.Contains(searchTerm, StringComparison.OrdinalIgnoreCase);
```
JeanPhilippeKernel: I'd prefer that we support more efficient search pattern algorithm here like KMP (Knuth-Morris-Pratt)

Panzerfaust/ViewModels/MainWindowViewModel.cs
```diff
@@ -18,27 +19,63 @@ namespace Panzerfaust.ViewModels
...
+
         public ReactiveCommand<Unit, Unit> CreateProjectCommand { get; }
         public Interaction<ProjectWindowViewModel, ProjectViewModel?> NewProjectDialog { get; } = new();
         public Interaction<MessageBoxWindowViewModel, bool> DeleteProjectInteraction { get; } = new();
 
+
         public MainWindowViewModel()
         {
+            var filterPredicate = this.WhenAnyValue(x => x.SearchText)
+                .Throttle(TimeSpan.FromMilliseconds(300), RxApp.MainThreadScheduler)
```
JeanPhilippeKernel: why the throttling here ?

JeanPhilippeKernel (changes requested):

Panzerfaust/ViewModels/MainWindowViewModel.cs
JeanPhilippeKernel: you can take an inspiration by following this impl : [Search pattern KMP](https://github.com/JeanPhilippeKernel/RendererEngine/blob/develop/Tetragrama/Helpers/SearchPatternAlgorithm.cpp)

AmanMenda: Oh okay, i'll read the implementation and adapt mine

Panzerfaust/ViewModels/MainWindowViewModel.cs
AmanMenda: The throttle here keep the "updated element" long enough for the changes to propagate. If multiple changes are made (user type a word for instance) within the specified time window (300ms), only the most recent one passes through (each previous changes are dropped and only the full word is taken into account in the filter. If user type "Flappy" for instance, i don't filter the 'f', 'fl', 'fla', etc.

Panzerfaust/ViewModels/MainWindowViewModel.cs
AmanMenda: I've managed to grasp the way the KMP algorithm works. Please keep me informed if anything need to be changed :) !

Panzerfaust/ViewModels/MainWindowViewModel.cs
JeanPhilippeKernel: With the new search pattern in place, am not sure we need any throttling anymore

JeanPhilippeKernel (changes requested):

Panzerfaust/ViewModels/MainWindowViewModel.cs
AmanMenda: You're right ! :)

JeanPhilippeKernel (approved):

PR #455, opened 2025-06-13
Code Restructure To Enable Linux and MacOS Builds, Plus Reduce Disk Usage and/or Build Times , Plus Help Readability of Source Files

MathewBensonCode:
- Move from github-submodule structure to use CMake's native dependency management facilities. (https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html). Ideally the dependencies are to be installed using native mechanisms like vcpkg on windows, the linux distribution package manager or homebrew for mac. CMake will find the installed versions as a first option but build the dependencies if not available in the Build Directory at Build time. The [linux_build](https://github.com/MathewBensonCode/RendererEngine/actions/runs/15634576278/job/44046460243) code action is an example of this working in linux.

- Restructure The #include statements to indicate the project source, this prevents confusion and is easier to read. It is also a consistent way to describe the included header and to reason about the include structure. This also is to pave the way to converting to modules

- Explicit Code file declaration in CMake as opposed to File Globbing which is only run once during build system generation. When creating a new file, it is suggested to specify it in the CMake File which will enhance tracking and ensure that it is picked up at build time due to changes in the CMake Files.

- Changed Bitmap.h to utf-8 from iso-8859

- Changed from MSVC style of pre-compiled headers, just using plain header includes for portability. Another benefit is storage savings as the pch file was taking up too much space(in my experience)

- I changed a few types and enums that we're throwing errors in g++ due to the typename and variable name being the same by adding "Enum" to the enum name and "Type" to the struct/class name. This may make the naming convention inconsistent, but it is something to be considered to be changed for all types so that there is a consistent naming convention similar to the vulkan style where there is a style for functions, types and constants

### DIFF
--- a/Resources/Shaders/infinite_grid.vert
+++ b/Resources/Shaders/infinite_grid.vert
@@ -13,7 +13,7 @@ Camera;
 
 void main()
 {
-    scaleFactor   = 80.0;
+    scaleFactor   = 50.0;
 
     vec3 posScale = pos * scaleFactor;
     uv            = posScale.xz;
--- a/Resources/Shaders/vertex_common.glsl
+++ b/Resources/Shaders/vertex_common.glsl
@@ -7,12 +7,14 @@ struct DrawVertex
 
 struct DrawData
 {
+    uint VertexOffset;
+    uint VertexCount;
+    uint IndexOffset;
+    uint IndexCount;
+    uint AllocationCount;
+    uint InstanceCount;
     uint TransformIndex;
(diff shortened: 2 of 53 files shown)


JeanPhilippeKernel:
Hi @MathewBensonCode, 
Thanks for your contribution and the PR

Your PR appears to introduce breaking changes and taking directions we aren't sure we want to follow for now.
I think you might want to split the PR for it to get reviewed and/or merged.

Thanks

PR #462, opened 2025-08-13
Enable CrossPlatform Builds and CI

MathewBensonCode:
# Changes made to enable cross platform builds

- Upgrade submodule dependencies to newer versions as upstream generally have fixed issues with them. For example assimp -Werror issues and GLM requiring old CMake Flags
- Rename variables that have the same name as the naming type as this wasn't working with GCC.
- Replicate Linux CI workflow using Ninja Generator for Linux
- Remove hard requirement for gcc11 and let the default compiler on the system to be used

### DIFF
--- a/.github/workflows/job-deploy-linux.yml
+++ b/.github/workflows/job-deploy-linux.yml
@@ -1,0 +1,29 @@
+# Deploy for linux
+#
+name: linux Deploy Workflow
+
+on:
+  workflow_call:
+    inputs:
+      targetFramework:
+        type: string
+        required: true
+      configuration:
+        type: string
+        default: 'Release'
+
+jobs:
+  deploy:
+    name: deploy-linux-${{ inputs.configuration }}
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          path: Result.Linux.x64.${{ inputs.configuration }}
+          name: Build-linux-Release
+
+      - name: Publish Artifacts
+        uses: actions/upload-artifact@v4
+        with:
+          name: linux-${{ inputs.configuration }}
+          path: Result.Linux.x64.${{inputs.configuration}}/Panzerfaust/${{ inputs.configuration }}/${{inputs.targetFramework}}/publish/
--- a/.github/workflows/linux-build.yml
+++ b/.github/workflows/linux-build.yml
@@ -1,59 +1,35 @@
-# @JeanPhilippeKernel : Disabled because we only support Windows as platform for now
-#
 name: ZEngine Linux Build
 
 on:
-  push:
-    branches: [ master ]
-  pull_request:
-    branches: [ master ]
+  workflow_call:
+    inputs:
+      targetFramework:
+        required: true
+        type: string
 
 jobs:
-  Linux-Build:
-    runs-on: ubuntu-latest
+  cmake-build:
     strategy:
       matrix:
-          buildConfiguration: [Debug, Release]
+        buildConfiguration: [Debug, Release]
+    uses: ./.github/workflows/job-cmakebuild-linux.yml
+    with:
+      configuration: ${{matrix.buildConfiguration}}
+      targetframework: ${{inputs.targetFramework}}
 
-    steps:
-    - name: Checkout repository
-      uses: actions/checkout@v2
+  test:
+    needs: cmake-build
+    strategy:
+      matrix:
+        testConfiguration: [Debug, Release]
+    uses: ./.github/workflows/job-test-linux.yml
+    with:
+      configuration: ${{matrix.testConfiguration}}
 
-#     - name: Checkout submodules
-#       run: git submodule update --init --recursive
+  deploy:
+    needs: test
+    uses: ./.github/workflows/job-deploy-linux.yml
+    with:
+      configuration: Release
+      targetframework: ${{inputs.targetFramework}}
 
-#     - name: Install development library
-#       run: sudo apt-get install libasound2-dev libgl1-mesa-dev libpulse-dev libdbus-1-dev libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev
-
-#     - name: Install CMake
-#       uses: jwlawson/actions-setup-cmake@v1.9
-#       with:
-#         cmake-version: '3.20.x'
-
-#     - name: Install pre-requisite packages.
-#       run: sudo apt-get install -y wget apt-transport-https software-properties-common
-
-#     - name: Download the Microsoft repository GPG keys
-#       run: wget -q https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb
-
-#     - name: Register the Microsoft repository GPG keys
-#       run: sudo dpkg -i packages-microsoft-prod.deb
-
-#     - name: Update the list of packages after we added packages.microsoft.com
-#       run: sudo apt-get update
-
-#     - name: Install PowerShell
-#       run: sudo apt-get install -y powershell
-
-#     - name: Add GCC Toolchain repository
-#       run: sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
-
-#     - name: Install GCC compiler
-#       run: sudo apt install -y gcc-11
-
-#     - name: Install G++ compiler
-#       run: sudo apt install -y g++-11
-
-#     - name: CMake Build
-#       run: .\Scripts\BuildEngine.ps1 -Configurations ${{matrix.buildConfiguration}}
-#       shell: pwsh
--- a/Scripts/BuildEngine.ps1
+++ b/Scripts/BuildEngine.ps1
@@ -113,7 +113,6 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
     # Check the system name
     if ($IsLinux) {
         $systemName = "Linux"
-        $cMakeGenerator
     }
     elseif ($IsMacOS) {
         $systemName = "Darwin"
@@ -131,7 +130,6 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
     [string]$BuildDirectoryName = "Result." + $systemName + "." + $architecture + "." + $BuildDirectoryNameExtension
     [string]$buildDirectoryPath = [IO.Path]::Combine($RepoRoot, $BuildDirectoryName)
     [string]$cMakeCacheVariableOverride = ""
-    [string]$cMakeGenerator = ""
 
     # Create build directory
     if (-Not (Test-Path $buildDirectoryPath)) {
@@ -171,11 +169,7 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
             $cMakeCacheVariableOverride += ' -DCMAKE_CONFIGURATION_TYPES=Debug;Release '
         }
         "Linux" {
-            $cMakeGenerator = "-G `"Unix Makefiles`""
-
-            # Set Linux build compiler
-            $env:CC = '/usr/bin/gcc-11'
-            $env:CXX = '/usr/bin/g++-11'
+            $cMakeGenerator = "-G `"Ninja`""
         }
         "Darwin" {
             $cMakeGenerator = "-G `"Xcode`""
--- a/Scripts/PostBuild.ps1
+++ b/Scripts/PostBuild.ps1
@@ -28,7 +28,7 @@ param (
     [string[]] $SystemName = 'Windows',
 
     [Parameter(HelpMessage="Architecture type to build, default to x64")]
-    [ValidateSet('win-x64', 'arm64', 'osx-x64', 'osx-arm64')]
+    [ValidateSet('win-x64', 'arm64', 'osx-x64', 'osx-arm64', 'linux-x64')]
     [string[]] $Architectures = 'win-x64',
 
     [Parameter(HelpMessage="Configuration type to build, default to Debug")]
@@ -103,7 +103,8 @@ $ContentsToProcess = @(
                     }
                 }
                 "Linux" {
-                    @{ From = "$OuputBuildDirectory\Obelisk\$Configurations"; To = "$OuputBuildDirectory\Panzerfaust\$Configurations\$TargetFramework\Editor"}
+                    @{ From = "$OuputBuildDirectory\Obelisk"; To = "$OuputBuildDirectory\Panzerfaust\$Configurations\$TargetFramework\Editor"}
+                    @{ From = "$OuputBuildDirectory\Obelisk"; To = "$OuputBuildDirectory\Panzerfaust\$Configurations\$TargetFramework\publish\Editor"}
                 }
                 Default {
                     throw 'This system is not supported'
--- a/Scripts/RunTests.ps1
+++ b/Scripts/RunTests.ps1
@@ -67,7 +67,9 @@ function RunTests {
         "Darwin" {
             $testExecutablePath = Join-Path $OutputBuildDirectory -ChildPath "ZEngine/tests/$Configuration/ZEngineTests"
         }
-        "Linux" {}
+        "Linux" {
+            $testExecutablePath = Join-Path $OutputBuildDirectory -ChildPath "ZEngine/tests/ZEngineTests"
+            }
         Default {
             throw 'This system is not supported'
         }
@@ -79,7 +81,7 @@ function RunTests {
         & $testExecutablePath
     }
     else {
-        Write-Warning "Test executable does not exist: $testExecutablePath"
+        Write-Error "Test executable does not exist: $testExecutablePath"
     }
 }
 
--- a/Scripts/Shared.ps1
+++ b/Scripts/Shared.ps1
@@ -189,7 +189,6 @@ function Find-ClangFormat () {
     $LLVMMaximumVersion = $repoConfiguration.Requirements.LLVM.MaximumVersion
 
     $candidates = @(
-        'clang-format'
         if ($IsMacOS) {
             $brewPrefixPath = Invoke-Expression -Command "& brew --prefix"
             Join-Path $brewPrefixPath -ChildPath 'opt/llvm/bin/clang-format'
@@ -197,6 +196,11 @@ function Find-ClangFormat () {
         if ($IsWindows) {
             Join-Path -Path $env:ProgramFiles -ChildPath 'LLVM\bin\clang-format.exe'
         }
+        if ($IsLinux) {
+            '/usr/bin/clang-format'
+            '/usr/bin/clang-format-20'
+        }
+        'clang-format'
     )
 
     foreach ($candidate in $candidates) {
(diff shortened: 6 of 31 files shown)


Scripts/BuildEngine.ps1
```diff
@@ -131,28 +116,26 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
...
         $Null = New-Item -ItemType Directory -Path $BuildDirectoryPath -ErrorAction SilentlyContinue
     }
 
     # Define CMake Generator arguments
-    $cMakeOptions = " -DCMAKE_SYSTEM_NAME=$systemName", " -DCMAKE_BUILD_TYPE=$configuration"
+    $cMakeOptions = " -DCMAKE_SYSTEM_NAME=$systemName", " -DCMAKE_BUILD_TYPE=$configuration", " -DBUILD_SHARED_LIBS=OFF"
     $submoduleCMakeOptions = @{
         'ENTT'      = @("-DENTT_INCLUDE_HEADERS=ON")
-        'SPDLOG'    = @("-DSPDLOG_BUILD_SHARED=OFF", "-DBUILD_STATIC_LIBS=ON", "-DSPDLOG_FMT_EXTERNAL=ON", "-DSPDLOG_FMT_EXTERNAL_HO=OFF");
-        'GLFW '     = @("-DGLFW_BUILD_DOCS=OFF", "-DGLFW_BUILD_EXAMPLES=OFF", "-DGLFW_INSTALL=OFF");
```
JeanPhilippeKernel: any reason why removing "-DGLFW_BUILD_EXAMPLES=OFF"  ?

Scripts/PostBuild.ps1
```diff
@@ -43,11 +43,7 @@ $ErrorActionPreference = "Stop"
 $TargetFramework = 'net8.0'
 
 [string]$RepoRoot = [IO.Path]::Combine($PSScriptRoot, "..")
-[string]$OuputBuildDirectory = If($IsWindows) {
-    [IO.Path]::Combine($RepoRoot, "Result.Windows.x64.MultiConfig")
-}  Else {
-    [IO.Path]::Combine($RepoRoot, "Result.$SystemName.x64.$Configurations")
-}
+[string]$OuputBuildDirectory =  [IO.Path]::Combine($RepoRoot, "../builds/RendererEngine")
```
JeanPhilippeKernel: Not sure this change 
the project has been configured to redirect the build output into a particular directory based on the platform - arch - and config

Scripts/Shared.ps1
```diff
@@ -185,8 +185,6 @@ function Find-GlslangValidator () {
 
 function Find-ClangFormat () {
     $repoConfiguration = Get-RepositoryConfiguration
-    $LLVMMinimumVersion = $repoConfiguration.Requirements.LLVM.Version
-    $LLVMMaximumVersion = $repoConfiguration.Requirements.LLVM.MaximumVersion
```
JeanPhilippeKernel: please revert -  LLVM has the tendency to introduce breaking change between each version -  and to avoid that we enforce sticking to specific version - which is configure in `repoConfiguration.json file`

Scripts/Shared.ps1
```diff
@@ -202,19 +200,16 @@ function Find-ClangFormat () {
...
-            if ((& $clangFormatCommand --version | Out-String) -match "clang-format version ([\d\.]*)") {
-                [Version] $clangFormatVersion = $Matches[1]
-                if ((CompareVersion $clangFormatVersion $LLVMMinimumVersion) -and (CompareVersion $LLVMMaximumVersion $clangFormatVersion)) {
                     return $clangFormatCommand.Source
-                }
-            }
         }
     }
 
-    throw "Failed to find clang-format min $LLVMMinimumVersion max $LLVMMaximumVersion. Tried: " + ($candidates -join ', ')
```
JeanPhilippeKernel: Same here - see my comment about the LLVM breaking change

Scripts/Shared.ps1
```diff
@@ -202,19 +200,16 @@ function Find-ClangFormat () {
...
 }
 
 function Setup-ShaderCCompilerTool () {
+    if($IsLinux) {
+        return
+    }
```
JeanPhilippeKernel: I believe ShaderC compiler has a version for linux -- @jnyfah can you check that ?

ZEngine/CMakeLists.txt
```diff
@@ -6,18 +6,5 @@ project (ZEngine
...
 
-# Compiler verification in Linux
-#
-if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-	if ((NOT ${CMAKE_CXX_COMPILER_VERSION} MATCHES "11.*.*") AND (${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang"))
-		message(FATAL_ERROR "CLANG compiler version should be 11.*.* !")
-	elseif ((NOT ${CMAKE_CXX_COMPILER_VERSION} MATCHES "11.*.*") AND (${CMAKE_CXX_COMPILER_ID} STREQUAL "GNU"))
-		message(FATAL_ERROR "GNU compiler version should be 11.*.* !")
-	endif ()
-endif ()
```
JeanPhilippeKernel, line 17: at the time I added this compiler verification - GCC 11 was the version that has a non-buggy support for coroutine?
do you ensure that all newer versions are good now? same for Clang

ZEngine/ZEngine/Rendering/Renderers/GraphicRenderer.cpp
```diff
@@ -15,7 +15,7 @@
 
 #define STB_IMAGE_WRITE_IMPLEMENTATION
 #define STB_IMAGE_RESIZE_IMPLEMENTATION
-#include <stb/stb_image_resize.h>
+#include <stb/deprecated/stb_image_resize.h>
```
JeanPhilippeKernel: any reason ?

ZEngine/ZEngine/Windows/Inputs/KeyCode.h
```diff
@@ -136,4 +137,9 @@ namespace ZEngine::Windows::Inputs
         MOUSE_BUTTON_8      = 7
 
     } GlfwKey;
+
+    inline auto format_as(GlfwKey f)
+    {
+        return fmt::underlying(f);
+    }
```
JeanPhilippeKernel: ??

ZEngine/ZEngine/CMakeLists.txt
```diff
@@ -67,16 +58,9 @@ target_compile_definitions (zEngineLib
 		YAML_CPP_STATIC_DEFINE
 )
 
-if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-	target_compile_definitions (zEngineLib PUBLIC VK_USE_PLATFORM_WIN32_KHR)
-endif()
```
JeanPhilippeKernel: any reason ??

ZEngine/ZEngine/CMakeLists.txt
```diff
@@ -67,16 +58,9 @@ target_compile_definitions (zEngineLib
...
-
-if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
-	target_link_libraries(zEngineLib PRIVATE stdc++fs)
-endif ()
```
JeanPhilippeKernel: same any reason ?

__externals/externals.cmake
```diff
@@ -1,60 +1,80 @@
...
-	target_link_libraries(imgui PUBLIC ${CMAKE_DL_LIBS})
+if(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
+    target_compile_definitions (ImGuizmo PUBLIC VK_USE_PLATFORM_WIN32_KHR)
 endif()
 
-add_library (imported::External_libs INTERFACE IMPORTED)
-add_library(imported::External_editorLibs INTERFACE IMPORTED)
+if (${CMAKE_SYSTEM_NAME} STREQUAL "Linux")
+    target_link_libraries(ImGuizmo PUBLIC stdc++fs)
+endif ()
```
JeanPhilippeKernel: No, ENABLE_VULKAN_VALIDATION_LAYER, VK_USE_PLATFORM_WIN32_KHR and stdc++fs are needed by zEngine lib , not ImGuizmo.  please revert

__externals/externals.cmake
```diff
@@ -1,60 +1,80 @@
...
+add_subdirectory (${EXTERNAL_DIR}/SPIRV-Cross)
+add_subdirectory (${EXTERNAL_DIR}/gtest)
+add_subdirectory (${EXTERNAL_DIR}/VulkanMemoryAllocator)
+add_subdirectory (${EXTERNAL_DIR}/tlsf)
+add_subdirectory (${EXTERNAL_DIR}/rapidhash)
+add_subdirectory (${EXTERNAL_DIR}/CLI11)
+add_subdirectory (${EXTERNAL_DIR}/nlohmann_json)
+
+add_library (ZEngineLibs INTERFACE)
+add_library(TetragramaLibs INTERFACE)
```
JeanPhilippeKernel: this naming is confusing - the engine lib name itself is zEnginelib

JeanPhilippeKernel (changes requested):

.clang-format
```diff
@@ -120,7 +120,7 @@ PenaltyBreakTemplateDeclaration: 100
 PenaltyReturnTypeOnItsOwnLine: 300
 PointerAlignment: Left
 ReflowComments: true
-SortIncludes: true
```
JeanPhilippeKernel: any reason ?

__externals/imgui/CMakeLists.txt
```diff
@@ -14,16 +14,20 @@ list (APPEND IMGUI_SOURCE
...
+add_library (imgui STATIC)
+
+target_sources(imgui PRIVATE ${IMGUI_SOURCE})
 
 # include directories
 #
 target_include_directories (imgui
     PUBLIC
         ./src
-        ./src/backends
```
JeanPhilippeKernel: we don't need src/backend anymore ?

JeanPhilippeKernel (changes requested):

Scripts/Shared.ps1
jnyfah: yes it does

jnyfah:
@MathewBensonCode I think to make it easier you can break this PR into smaller PR's that focus on one thing at a time 
maybe first work on 
1. adding linux support and ensuring that the panzerfaust/launcher and the editor works correctly 
2. Once the Linux support PR is merged you can work on adding linux workflows
3. finally fix the submodule 

that way its not overwhelming and things are fixed one at a time 
let me know if you have any questions or issues

.github/workflows/job-deploy-linux.yml
```diff
@@ -0,0 +1,29 @@
...
+        default: 'Release'
+
+jobs:
+  deploy:
+    name: deploy-linux-${{ inputs.configuration }}
+    runs-on: ubuntu-24.04
+    steps:
+      - uses: actions/download-artifact@v4
+        with:
+          path: Result.Darwin.x64.${{ inputs.configuration }}
```
JeanPhilippeKernel:
```suggestion
          path: Result.Linux.x64.${{ inputs.configuration }}
```

.github/workflows/linux-build.yml
```diff
@@ -1,59 +1,37 @@
-# @JeanPhilippeKernel : Disabled because we only support Windows as platform for now
+# @JeanPhilippeKernel 
 #
```
JeanPhilippeKernel: we can drop this comment

JeanPhilippeKernel (changes requested):

Scripts/BuildEngine.ps1
MathewBensonCode: I read the upstream cmake file and found it was being set by default so I was removing the redundant option.

https://github.com/glfw/glfw/blob/master/CMakeLists.txt

Scripts/PostBuild.ps1
MathewBensonCode: will revert this, was left there accidentally.

Scripts/Shared.ps1
MathewBensonCode: I think we can give this a bit more thought as limiting the version will limit portability and flexibility. 

I was envisioning the ability to have the system build on a wide variety of configurations.

Perhaps testing can be done on the current stable version and have that set as the maximum.

ZEngine/CMakeLists.txt
MathewBensonCode: I think its safe to say that this is ok. The CI version of ubuntu uses version 13, and I have tested 15 and 16(trunk).

I think it would only break with older compilers which should be quite rare.

ZEngine/ZEngine/Rendering/Renderers/GraphicRenderer.cpp
MathewBensonCode: the upstream project moved this header to the deprecated folder as there is a newer version that would need a code rewrite to match.

ZEngine/ZEngine/Windows/Inputs/KeyCode.h
MathewBensonCode: There was a compilation error when trying to pass the enum glfwkey to the fmt lib. This is to tell fmt to use the underlying type of the enum as the formating type.

__externals/externals.cmake
MathewBensonCode: I had meant to put this in the imgui target as it is the one that pulls in vulkan.

The idea is that imgui pulls in glfw and vulkan, with their definitions. Then Imguizmo pulls in imgui and finally ZEngine pulls in ImGuizmo.

__externals/externals.cmake
MathewBensonCode: I agree. Perhaps I could rename them as ZEngineLibDependencies or something similar. 

In the same vein we should consider using a different name for the imported::*** versions.

Scripts/Shared.ps1
MathewBensonCode: I'll look into this. At a glance I saw that the dependency was being pulled in from an archive which I assumed to be yours so I was testing with the system installed version.

.clang-format
MathewBensonCode: There was a compilation issue that would come up after running clang-format in ImGUIRenderer.cpp where the ImGuizmo header when included before the imgui header would break as the imgui headers need to come in first.

__externals/imgui/CMakeLists.txt
MathewBensonCode: The backends directory will be specified in the #include directive in the source files.

Scripts/Shared.ps1
JeanPhilippeKernel: I believe we are already on the latest stable version: 
the config for external tool like llvm here is: 
https://github.com/JeanPhilippeKernel/RendererEngine/blob/develop/repoConfiguration.json

MathewBensonCode:
I have tried to capture the requested changes in a single commit which we can now look into splitting up into smaller PRs as necessary.

I found it a little tricky to split up because the changes were related but we can now discuss this further and you can let me know what you have in mind.

Scripts/Shared.ps1
MathewBensonCode: I see. I made some changes to the CI workflow to ensure the linux build can find and use clang-format 20 on ubuntu and some other distributions as well by adding options tp the searched paths.

ZEngine/ZEngine/Windows/Inputs/KeyCode.h
```diff
@@ -136,4 +137,10 @@ namespace ZEngine::Windows::Inputs
         MOUSE_BUTTON_8      = 7
 
     } GlfwKey;
+
+    inline std::ostream& operator<<(std::ostream& stream, GlfwKeyCode f)
+    {
+        stream << static_cast<int32_t>(f);
+        return stream;
+    }
```
JeanPhilippeKernel, line 145: any reason for this overload ?

Scripts/RunTests.ps1
```diff
@@ -79,11 +81,11 @@ function RunTests {
...
     else {
-        Write-Warning "Test executable does not exist: $testExecutablePath"
+        Write-Error "Test executable does not exist: $testExecutablePath"
     }
 }
 
 # Run tests for each configuration
 foreach ($config in $Configurations) {
-    RunTests -Configuration $config
+RunTests -Configuration $config
```
JeanPhilippeKernel: revert the space

Scripts/Shared.ps1
```diff
@@ -189,14 +189,18 @@ function Find-ClangFormat () {
     $LLVMMaximumVersion = $repoConfiguration.Requirements.LLVM.MaximumVersion
 
     $candidates = @(
-        'clang-format'
```
JeanPhilippeKernel, line 192: revert - it should be the first candidate before looking the specific path

ZEngine/CMakeLists.txt
JeanPhilippeKernel: as you said, it will break with older versions -  let's enforce that by telling the developer to upgrade its compiler version.

The project needs C++20 and coroutine support
for Clang  it should be minimum v11 and higher
for GCC it should be minimum v10 or 11  and higher

JeanPhilippeKernel (changes requested):

.clang-format
JeanPhilippeKernel: we can simply put that header in a formatting off section `// clang-format off ..... // clang-format on`
see example of usage here : https://github.com/JeanPhilippeKernel/RendererEngine/blob/develop/ZEngine/ZEngine/Hardwares/VulkanDevice.h

ZEngine/ZEngine/Windows/Inputs/KeyCode.h
MathewBensonCode: some compilers or versions of fmt have trouble formating enums. On newer versions there is a `format_as` function, but as We are not upgrading the submodules the `ostream <<` is the alternative.

https://github.com/fmtlib/fmt/issues/391

Scripts/Shared.ps1
MathewBensonCode: Doing it this way is important where clang-format version 20 is not the default, for example in the CI build where we are specifically installing it as well as on a system like mine where the path always selects the newer version first. If some other system specific version or path comes up in testing it would be easy to just add that to the search options

JeanPhilippeKernel: I see - alright then

ZEngine/ZEngine/Windows/Inputs/KeyCode.h
JeanPhilippeKernel: interesting - can you help creating an issue to track the upgrade of fmtlib

ZEngine/CMakeLists.txt
MathewBensonCode: This is actually being enforced by the CMAKE_CXX_STANDARD variable. If the compiler does not support c++20,  then the process ends here. CMake is aware of the versions and support so we do not have to do it manually.

https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html

Scripts/BuildEngine.ps1
```diff
@@ -171,11 +169,7 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
             $cMakeCacheVariableOverride += ' -DCMAKE_CONFIGURATION_TYPES=Debug;Release '
         }
         "Linux" {
-            $cMakeGenerator = "-G `"Unix Makefiles`""
-
-            # Set Linux build compiler
-            $env:CC = '/usr/bin/gcc-11'
-            $env:CXX = '/usr/bin/g++-11'
+            $cMakeGenerator = "-G `"Ninja`""
```
JeanPhilippeKernel, line 172: we want to update the README to provide installation steps for Ninja  on linux

Ideally we might want to automate that installation so devs wont need to worry about it.  we can do that in follow-up PR - please create an issue to track it

MathewBensonCode: done

jnyfah (changes requested):
i dont know why i cant comment on the file, but `shared.ps1` add linux to the `GlslangValidatorCandidates` line 161

`
    $GlslangValidatorCandidates = @(
        'glslangValidator'

        if ($IsWindows) {
            Join-Path -Path $shaderCCompilerPath -ChildPath "\bin\glslangValidator.exe" # On Windows, the pipeline build might pick up this option...
        }
        if ($IsMacOS) {
            Join-Path -Path $shaderCCompilerPath -ChildPath "\bin\glslangValidator" # On macOS, the pipeline build might pick up this option...
        }
        if ($IsLinux) {
           Join-Path -Path $shaderCCompilerPath -ChildPath "\bin\glslangValidator"
        }
    )
`

jnyfah (approved):

jnyfah (changes requested):
everything looks good to me, just make the changes on the glslangcandidate

JeanPhilippeKernel (approved):

jnyfah (approved):

PR #468, opened 2025-08-25
Fix incorrect pre-compiled-header usage

MathewBensonCode:
- Remove `#include <pch.h>` directive from source and header files as the benefits of using precompiled headers with cmake comes in not adding that compilation burder per translation unit.
- The `target_precompile_headers()` cmake command was being used correctly, but the include statement negates the benefits

  https://cmake.org/cmake/help/latest/command/target_precompile_headers.html


JeanPhilippeKernel:
Thanks @MathewBensonCode for this PR !!

JeanPhilippeKernel (approved):

PR #471, opened 2025-09-01
Removed GLM Library

jnyfah:


JeanPhilippeKernel (approved):
Thanks for the hard work on dropping glm

PR #480, opened 2025-11-12
Introduce CMake fetchcontent and CMake Install Workflow

MathewBensonCode:
# FetchContent Package Management
- Here is an idea for migrating from the git module based dependency management to use FetchContent. 
- The packages are still built by the build system in order to preserve the versions. 
- This is done by avoiding the use of the Find_Package method that would use what is installed on the system. 
- This makes it virtually the same.

# CMake Install
 - This replaces the same functionality that was in the PostBuild.ps1 script
 - Used the cmake install features to ensure consistent installation location, which is canonical i.e. binaries in the bin directory and libraries in the lib directory
 - As I was debugging I found that there was some inconsistency with the paths to the shaders in Panzerfaust vs Obelisk. e.g. if you ran obelisk directly without going through Panzerfaust the built in shader location wasn't the same.

# SwapChain Image Sizes
- Was getting a vulkan error that the system was running out of memory.
- Traced this to the size being passed to the SwapChainCreateFunction.
- Initialize SwapchainImages with the size parameter added as the creation of the swap chains was happening successfully but the size was not getting updated giving an index error.

# Wayland on Linux
- Fix error on linux where Wayland was not being initialized properly. Wayland is the default for glfw in version 3.4 and in many desktop environments. GLFW_EXPOSE_NATIVE_* macro was defined for windows but not for the other Operating Systems


MathewBensonCode:
@JeanPhilippeKernel @jnyfah I'm not sure why this check is failing, however, I would like to request you to try the [build artifacts](https://github.com/MathewBensonCode/RendererEngine/actions/runs/19404060021#artifacts) on your systems. 

The main thing with the install workflow is to move what was in the Panzerfaust/net8.0..../publish folder is not going to the bin folder with the vulkan loader library in the lib folder.

This means that downloading the release builds should be able to run on a user system easily. 

The program doesn't run because of the "Shader::CreateDescriptorLayout()" code runs out of vram when loading up the shaders.

However, I believe that the "not running on mac and linux" issue may be resolved by this.

JeanPhilippeKernel:
Hey @MathewBenson 
Thanks a lot for the PR - I re-ran the pipeline and it's green now -

JeanPhilippeKernel:
my first observation will be to extract the vulkan related issue in a seperate PR  - and as heads-up there is an on-going effort to fix the issue here : https://github.com/JeanPhilippeKernel/RendererEngine/pull/479

ZEngine/ZEngine/Rendering/Specifications/ShaderSpecification.h
```diff
@@ -38,14 +38,14 @@ namespace ZEngine::Rendering::Specifications
         uint32_t         Set{0xFFFFFFFF};
         uint32_t         Binding{0xFFFFFFFF};
         uint32_t         Count{1};
-        std::string      Name;
+        const char*      Name;
```
JeanPhilippeKernel: i think we have a typedef `cstring` - you might want to use that

ZEngine/ZEngine/Rendering/Specifications/ShaderSpecification.h
```diff
@@ -38,14 +38,14 @@ namespace ZEngine::Rendering::Specifications
...
         DescriptorType   DescriptorTypeValue;
         ShaderStageFlags Flags;
     };
 
     struct PushConstantSpecification
     {
-        std::string      Name;
+        const char*      Name;
```
JeanPhilippeKernel: same comment about the typedef `cstring`

CMakePresets.json
```diff
@@ -20,7 +25,7 @@
 
                 "GLFW_BUILD_DOCS": "OFF",
                 "GLFW_BUILD_EXAMPLES": "OFF",
-                "GLFW_INSTALL": "OFF",
+                "GLFW_BUILD_WAYLAND": "ON",
```
JeanPhilippeKernel: I'd prefer that , on windows and macos preset , we explicitly OFF it , even though it might be ignored

ZEngine/ZEngine/CMakeLists.txt
```diff
@@ -55,32 +49,35 @@ target_include_directories (zEngineLib
 		./Windows/Layers
 		./Managers
 		./Serializers
-		${EXTERNAL_DIR}
-		${EXTERNAL_INCLUDE_DIRS}
 )
 
 target_precompile_headers(zEngineLib PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/pch.h)
 
+target_compile_definitions(zEngineLib PUBLIC GLM_ENABLE_EXPERIMENTAL)
```
JeanPhilippeKernel: we now have our internal math lib, so i believe we can get rid of GLM - cc @jnyfah since you're the lead on it

JeanPhilippeKernel (changes requested):

dependencies.cmake
```diff
@@ -1,3 +1,282 @@
...
+          ${IMGUIDIR}/backends/imgui_impl_vulkan.cpp)
+
+      target_include_directories(imgui 
+          PUBLIC ${FETCHCONTENT_BASE_DIR}
+          PUBLIC ${FETCHCONTENT_BASE_DIR}/imgui
+      )
+
+target_compile_definitions(imgui PUBLIC GLFW_INCLUDE_VULKAN IMGUI_DEFINE_MATH_OPERATORS)
+
+target_link_libraries(imgui PUBLIC glfw Vulkan::Headers Vulkan::Loader)
```
JeanPhilippeKernel: This is problematic ! imgui should consume `Vulkan::Headers` and `Vulkan::Loader` from zEngineLibs

imgui (and its plugin libs like imguizmo) is an optional lib for zEngineLibs. 
it just helps for GUI components, from Tetragramma (The editor) and nothing more - and there is a big chance that we might want to have our internal GUI lib and drop it.
So we want the fexibility to replace it anytime 

Ideally, it should be moved to Tetagramma - that will require some refactoring from the Rendering section -

JeanPhilippeKernel (changes requested):

Obelisk/CMakeLists.txt
```diff
@@ -44,8 +42,20 @@ target_compile_definitions(${TARGET_NAME}
         _UNICODE
 )
 
-target_link_libraries(${TARGET_NAME} PRIVATE
-    zEngineLib
```
JeanPhilippeKernel, line 48: not sure i fully understand the removal here

Obelisk/EntryPoint.cpp
```diff
@@ -1,4 +1,5 @@
 #include <CLI/CLI.hpp>
+#include <Editor.h>
```
JeanPhilippeKernel: there is already `#include <Tetragrama/Editor.h>` below

JeanPhilippeKernel (changes requested):

CMakeLists.txt
```diff
@@ -85,32 +64,4 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
...
-if (NOT LAUNCHER_ONLY)
-	add_dependencies(AssembleContent zEngineLib tetragrama Obelisk)
-endif ()
-
-if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-	add_dependencies(AssembleContent Panzerfaust-WinBuild)
-else ()
-	add_dependencies(AssembleContent Panzerfaust)
-endif ()
+include(CPack)
```
JeanPhilippeKernel: maybe you want to move it close to other include(..) at the top

CMakePresets.json
```diff
@@ -40,13 +45,12 @@
...
                 "YAML_BUILD_SHARED_LIBS": "OFF",
 
-                "USE_MASM": "OFF",
                 "USE_GAS": "OFF",
 
                 "SPIRV_SKIP_EXECUTABLES": "ON",
                 "SPIRV_SKIP_TESTS": "ON",
                 "GLSLANG_ENABLE_INSTALL": "ON",
-                "ALLOW_EXTERNAL_SPIRV_TOOLS": "OFF",
+                "ALLOW_EXTERNAL_SPIRV_TOOLS": "ON",
```
JeanPhilippeKernel, line 52: We want to OFF this because we want to use the one we are compiling during the build and avoid any version mismatching with the system -  any reason why this ON ?

Scripts/BuildEngine.ps1
```diff
@@ -142,17 +132,38 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
...
         $buildArguments = "--build --preset $configName"
 
         $buildProcess = Start-Process $cMakeProgram -ArgumentList $buildArguments -NoNewWindow -PassThru
 
-        # Grab the process handle. When using `-NoNewWindow`, retrieving the ExitCode can return null once the process
-        # has exited. See:
-        # https://stackoverflow.com/questions/44057728/start-process-system-diagnostics-process-exitcode-is-null-with-nonewwindow
+# Grab the process handle. When using `-NoNewWindow`, retrieving the ExitCode can return null once the process
+# has exited. See:
+# https://stackoverflow.com/questions/44057728/start-process-system-diagnostics-process-exitcode-is-null-with-nonewwindow
```
JeanPhilippeKernel: fix formatting

JeanPhilippeKernel (changes requested):

CMakeLists.txt
```diff
@@ -85,32 +64,4 @@ if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
...
-
-if (NOT LAUNCHER_ONLY)
-	add_dependencies(AssembleContent zEngineLib tetragrama Obelisk)
-endif ()
-
-if (${CMAKE_SYSTEM_NAME} STREQUAL "Windows")
-	add_dependencies(AssembleContent Panzerfaust-WinBuild)
-else ()
-	add_dependencies(AssembleContent Panzerfaust)
-endif ()
```
JeanPhilippeKernel, line 116: we can't just drop all this - we'd to rethink this part - remember : we have `LAUNCHER_ONLY` option which helps to build only Panzerfaust - the launcher and skip all the rest (zEngine, Obelisk, Tetragrama)

JeanPhilippeKernel (changes requested):

ZEngine/ZEngine/Rendering/Specifications/ShaderSpecification.h
MathewBensonCode: same as above

CMakePresets.json
MathewBensonCode: yes. will move this to the linux specific part. I spotted it after.

ZEngine/ZEngine/CMakeLists.txt
MathewBensonCode: I agree. This was erroneosly included during a merge. will remove it

CMakeLists.txt
MathewBensonCode: This is part of the migration to using cmake install. The Functionality is handled in line 41 of the CMakeLists.txt file, where if we only want the launcher then the targets for zEngine, Obelisk and Tetragrama are not created. 

There is a modification in the Panzerfaust/CMakeLists.txt file that uses the install cmake command to install the Panzerfaust files to the bin/ folder.

The Selection is handled in the BuildEngine.ps1 as before. I did try not to break this workflow. Kindly give the `Scripts/BuildEngine.ps1 -LauncherOnly ` command and let me know if it works on your end.

Ideally it should do the same thing that was being done in the PostBuild.ps1 and the section of the CMakeLists.txt file that you just quoted

dependencies.cmake
MathewBensonCode: As imgui and ImGuizmo do not have CMakeLists.txt files to create targets, I went with the option of creating them as separate targets as opposed to including their source files in the zEngine library.

I am using the transitive property of CMake Dependencies to pass the Vulkan and glfw from imgui -> Imguizmo -> ExternalLib ->zEngineLib. The idea was to reduce repetition. 

I could make the dependency to be PRIVATE and specify them as "Build Requirements" for imgui and currently for zEngineLib.

This way imgui will have `target_link_libraries(imgui PRIVATE Vulkan::Headers....`
and zEngine will have `target_link_libraries(External_Libs PUBLIC Vulkan::Headers.....`

Obelisk/CMakeLists.txt
MathewBensonCode: Just as above the zEngineLib is transitively availed to Obelisk through Tetragrama.

CMakePresets.json
MathewBensonCode: The "ALLOW_EXTERNAL_SPIRV_TOOLS" is a CMake Property on GSLang that makes it to build its on SPIRV-Tools if they are not detected by CMake.

As we are already building SPIRV-Tools as an option,, we are telling it not to build its own version.

There is actually another potential repetition of this because of the step of downloading the compiler tools in the build step of the Powershell scripts. We could alleviate that download, which is like 1GB by making the shader compilation be handled as part of the CMake Processing where we use the binaries that we compile when generating GSLang Target.

JeanPhilippeKernel: > We could alleviate that download, which is like 1GB by making the shader compilation be handled as part of the CMake Processing where we use the binaries that we compile when generating GSLang Target.

Interesting,
we can explore this in separate PR - please create an issue to track that investigation

CMakeLists.txt
JeanPhilippeKernel: that's good ! yeah it works- thanks

dependencies.cmake
JeanPhilippeKernel: Sounds good to me  !

CMakePresets.json
MathewBensonCode: Created issue #481 to track. Coincidentally, the CI is now failing because the upstream ShaderC URLs are broken and/or the requested versions are unavailable on the server

JeanPhilippeKernel: What a timing ! 😅😂

JeanPhilippeKernel:
Hey @MathewBensonCode  any update on this ?

MathewBensonCode:
I believe I tackled the issues you raised up the last time but I guess we got side-tracked by the ShaderC compiler matter.

I did try to break down some of the changes that we're incorporated in this, through the smaller pull requests that we have worked on. 

I was waiting for your feedback on the matter of the "install" path modifications. 

I was also considering creating a separate pull request with just the fetch_content dependency management change then another with the install path change. 

The main aim of this "omnibus" pull request was mainly as a proof of concept for the workflow that you could have a taste test.

JeanPhilippeKernel:
I see - 
I think we good to have the fetch_content and the install path in this Pr but feel free to split them in two PRs.

But generally I'm okay with the new install paths

MathewBensonCode:
After trying to split the PRs, I remembered why I had done it in this way(it has been some time).

The change of dependencies, meant changes in the install paths and consequently the CI builds. So it just made sense to make the full switch.

Also, on the matter of transitive dependencies e.g. zEngine -> Tetragrama -> Obelisk, One VERY important reason to make the dependencies transitive is Header files that include other Header Files. If you have the dependencies as `private` then when you encounter a header file that includes another header file from a dependent project, things break in frustrating ways.

JeanPhilippeKernel (approved):

PR #485, opened 2026-01-15
Remove LLVM 20 limitation

### LINKED ISSUE #484: Remove LLVM 20 limitation, LLVM 21 current default on many platforms
There is a limitation in the repoConfiguration.json file limiting the Version of LLVM to version 20. This is due to breaking changes when formatting with clang-format.

The current version on many platforms is now LLVM 21 so trying to build on these systems using the Build Scripts fails due to this requirement.

The issue arises due to the `operator bool()` function in two files in particular `CoroutineScheduler.h` and `VulkanDevice.h` which looks like an identifier to the current custom `clang-format` rules.

MathewBensonCode:
- clang-format ignore on CoroutineScheduler.h and VulkanDevice.h for operator bool()
- update ubuntu CI workflow to use version 21.

Closes #484

### DIFF
--- a/.github/workflows/job-cmakebuild-linux.yml
+++ b/.github/workflows/job-cmakebuild-linux.yml
@@ -25,16 +25,17 @@ jobs:
     - name: Register the Microsoft repository GPG keys
       run: sudo dpkg -i packages-microsoft-prod.deb
 
-    - name: Install LLVM-20
+    - name: Install LLVM-22
       run: |
           wget https://apt.llvm.org/llvm.sh
           chmod +x llvm.sh
-          sudo ./llvm.sh 20
+          sudo ./llvm.sh 22
 
-    - name: Update the list of packages after we added packages.microsoft.com
+    - name: Update the list of packages
       run: sudo apt-get update
-    - name: Install development library
-      run: sudo apt-get install libasound2-dev libgl1-mesa-dev libpulse-dev libdbus-1-dev libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev cmake apt-transport-https software-properties-common gcc g++ powershell clang-format-20 libwayland-dev libxkbcommon-dev dotnet-sdk-8.0
+
+    - name: Install Dependencies
+      run: sudo apt-get install libasound2-dev libgl1-mesa-dev libpulse-dev libdbus-1-dev libx11-dev libxcursor-dev libxext-dev libxi-dev libxinerama-dev libxrandr-dev libxss-dev libxt-dev libxxf86vm-dev cmake apt-transport-https software-properties-common gcc g++ powershell clang-format-22 glslang-tools libwayland-dev libxkbcommon-dev
 
     - name: CMake Build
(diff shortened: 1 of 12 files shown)


Scripts/Shared.ps1
```diff
@@ -201,6 +201,7 @@ function Find-ClangFormat () {
         }
         if ($IsLinux) {
             '/usr/bin/clang-format'
+            '/usr/bin/clang-format-21'
             '/usr/bin/clang-format-20'
```
JeanPhilippeKernel, line 206: That means we want to remove the clang-format-20 from candidate list

.github/workflows/job-cmakebuild-linux.yml
```diff
@@ -25,17 +25,17 @@ jobs:
     - name: Register the Microsoft repository GPG keys
       run: sudo dpkg -i packages-microsoft-prod.deb
 
-    - name: Install LLVM-20
+    - name: Install LLVM-21
```
JeanPhilippeKernel: if we have to upgrade to LLVM-21 , that means we need to ensure that all platform we support , as well as pipeline agents have it available.

Also Readme need to be updated for contributors to be aware

JeanPhilippeKernel (changes requested):

Scripts/Shared.ps1
MathewBensonCode: I thought to leave it in as a precaution incase someone is still on the older platform. The main idea was to remove the restriction on newer versions but that compatibility with the older version is retained as specified in repoConfig.json

.github/workflows/job-cmakebuild-linux.yml
MathewBensonCode: I put this change in the CI as a proof of concept that version 21 does work as well as version 20.

JeanPhilippeKernel: Sure ! -
what I mean is that we need to propage the change to windows and macos as well

MathewBensonCode: Windows actually prompted the change as the default version moved to version 21. This is on developer machines. I used winget and it tracks the current stable version. I believe the same should be true for macOS.

Ubuntu and debian tend to be conservative so are always behind in their software versions, hence the need to specifically install version 20 and above. I think I recall that the ubuntu CI build would not work out of the box without this  specific installation as the maximum they had available was version 19.

It may be possible to lower the requirement in the repoConfiguration.json to earlier versions and not have to bother with constraining user versions as it should just work everywhere...(hopefully).

MathewBensonCode:
Updated this to support llvm version 22, which is now the current stable version. 

I also all the CIs including the clang-format steps to use clang-format version 22. 

I isolated the few areas which have issues in the newer versions by excluding them using `// clang-format off` calls.

This means that we should be able to work with any version from 20 upto 22, and even older(but haven't tested those)

JeanPhilippeKernel:
the spreading usage of `// clang-format off` proves that we might need soon to raise the bar of minimal version of clang-format we want to support - let's track it with an issue for later.

JeanPhilippeKernel (approved):

PR #492, opened 2026-01-19
Ensuring Update-After-Bind is supporting by the Device

### LINKED ISSUE #355: Make the Engine run on Linux Debian
Since the migration from OpenGL to Vulkan API, the support on Linux Debian has been left behind in favor Windows platforms
Now we fully landed and operable on Linux Debian, it's time to give focus on Linux Debian and do the port as well.

### LINKED ISSUE #493: Shader Creation Failure - Running out of VRAM when Creating DescriptorPool
When creating the shaders, one of the Shaders, particularly the g_buffer Shader, a very large amount of VRAM is occupied. In consequent shader creation, the process fails at the VK_ERROR_OUT_OF_DEVICE_MEMORY error.

This has been tested in both windows and linux and on intel_gpu and amd_apu and the error is now the same.

JeanPhilippeKernel:
In this PR, I revisited the process of Shader source reflexion to generate DescriptorSet and Layout. 

I introduced the TextureArray as global binding across all shaders instead of per- shader.

I also revisited the logic around DescriptorPool size creation, which was consumed large resources than needed.

Then, I hardened the detection  and support of Bindless resource of the Device so we can switch to a fallback option in case of no support.

Close #493


JeanPhilippeKernel:
@MathewBensonCode It will be great if you can test the linux artificat of this PR :)

MathewBensonCode:
Interesting you should ask this now as I was working on this yesterday.

It works fine and the previous issue is not present.

A new issue actually arises because of the change in Swapchain Image count affecting the commandbuffer manager and other related code. 

I Wasn't sure how to raise up this new issue when this pr wasn't yet merged. So I was going to wait. 

check out the `afterbind` branch on my repo to see a proposed solution to the new issue.

```

JeanPhilippeKernel:
> check out the `afterbind` branch on my repo to see a proposed solution to the new issue.

Interesting ! 
I got Zero warning on macOS 🤦🏽‍♂️ ! 

I think you can open a PR against this branch, so we can iterate on your proposal - just checked and It seems correct but there are few adjustments.

JeanPhilippeKernel:
@MathewBensonCode It'b be great if you can make a final testing on linux for this PR artificat - We should be pretty solid  now on all platforms

MathewBensonCode:
Linux is showing the UI as expected. Able to launch Obelisk through the Panzerfaust UI.

JeanPhilippeKernel:
On Windows, i'm observing a crash due to dangling pointers in Shader and RenderPass classes -  mostly on how we manipulate/process string

MathewBensonCode:
I tried this on windows yesterday and it worked ok.

jnyfah:
merging this since its confirmed working

jnyfah (approved):

PR #495, opened 2026-01-28
Mitigate undesired unsynchronized effect of changing SwapchainImage Count

### LINKED ISSUE #493: Shader Creation Failure - Running out of VRAM when Creating DescriptorPool

MathewBensonCode:
When creating the Swapchain we check with the device for the supported value. If this value is higher than the pre-configured values, structures created with the initial values are now out of sync, leading to crashes.

This approach attemps to create those structures after getting the supported Swapchain image count

### DIFF
--- a/ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
+++ b/ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
@@ -427,8 +427,6 @@ namespace ZEngine::Hardwares
 
         CreateSwapchain();
 
-        EnqueuedCommandbuffers.init(Arena, m_buffer_manager.TotalCommandBufferCount, m_buffer_manager.TotalCommandBufferCount);
-
         SwapchainRenderCompleteSemaphores.init(Arena, SwapchainImageCount, SwapchainImageCount);
         SwapchainAcquiredSemaphores.init(Arena, SwapchainImageCount, SwapchainImageCount);
         SwapchainSignalFences.init(Arena, SwapchainImageCount, SwapchainImageCount);
@@ -1242,7 +1240,8 @@ namespace ZEngine::Hardwares
             SwapchainFramebuffers.init(Arena, SwapchainImageCount, SwapchainImageCount);
         }
 
-        m_buffer_manager.Initialize(this, SwapchainImageCount);
+        m_buffer_manager.Initialize(this);
+        EnqueuedCommandbuffers.init(Arena, m_buffer_manager.TotalCommandBufferCount, m_buffer_manager.TotalCommandBufferCount);
 
         scratch                        = ZGetScratch(Arena);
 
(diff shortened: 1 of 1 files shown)


JeanPhilippeKernel:
I believe, addressing the conflict should reduce the file change and keep your changes on `CommandBufferManager`

ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
```diff
@@ -1242,6 +1242,8 @@ namespace ZEngine::Hardwares
             SwapchainFramebuffers.init(Arena, SwapchainImageCount, SwapchainImageCount);
         }
 
+        m_buffer_manager.Initialize(this, SwapchainImageCount);
```
JeanPhilippeKernel: Since we pass the Device to `Initialize(this)` we can simplify Initialize signature by dropping the params : `swapchain_image_count`
And internally to `Initialize(...)` we can read `SwapchainImageCount` from Device

See here : https://github.com/JeanPhilippeKernel/RendererEngine/blob/107b917bdf02b106a3184176052f675a4093154c/ZEngine/ZEngine/Hardwares/VulkanDevice.cpp#L1870

https://github.com/JeanPhilippeKernel/RendererEngine/blob/107b917bdf02b106a3184176052f675a4093154c/ZEngine/ZEngine/Hardwares/VulkanDevice.cpp#L1873

ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
```diff
@@ -428,6 +425,10 @@ namespace ZEngine::Hardwares
         PreviousFrameIndex                                    = 0;
         CurrentFrameIndex                                     = 0;
 
+        CreateSwapchain();
+
+        EnqueuedCommandbuffers.init(Arena, m_buffer_manager.TotalCommandBufferCount, m_buffer_manager.TotalCommandBufferCount);
```
JeanPhilippeKernel: This needs to be called right after  `m_buffer_manager.Initialize(...)` to allow `TotalCommandBufferCount` having the right value

JeanPhilippeKernel (changes requested):

ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
JeanPhilippeKernel: for more clarity, I think you can move your whole fix right before this line : 
`AsyncResLoader->Initialize(this);`

MathewBensonCode:
I would like to confirm that as at the last commit, the application is running successfully on linux. Its actually the first time I have seen the UI running since I started looking at this project a few years ago.

i'll take a look at the changes requested a little later.

ZEngine/ZEngine/Hardwares/VulkanDevice.cpp
MathewBensonCode: Just pushed a new commit with the requested change on the function parameter. 

I was however, unable to move the code to the requested location as the commandbuffermanager is needed from within the CreateSwapChain here https://github.com/JeanPhilippeKernel/RendererEngine/pull/495/changes/f45be5c0bc6df148a94824d495a2668830239f37#diff-c13ca6ac82de9f04347faeea8859a0e7bebfa7b5e2a38698aa35eae8d269bb13L1254

Also the lines just after the call to CreateSwapchain() all require the corrected SwapchainImageCount, So it wasn't possible to move the fix down to the requested location

PR #496, opened 2026-01-29
Missing change to previous commit

MathewBensonCode:
I missed out the header in the final commit to #495.

This is actually related to #485 as I generally have to keep reverting the changes made to the `VulkanHeader.h` and `CoroutineScheduler.h` files due to the formatting in the main script.

The changes in #485 are beneficial on any developer machine with the newer llvm versions as we put the `clang-format` guard on the problematic lines.


JeanPhilippeKernel:
ha - I already fixed it in the PR with other bunch of edges cases -

PR #497, opened 2026-01-30
fix: Header Re-Organization for Consistency and Readability

MathewBensonCode:
- Provide a consistent taxonomy for headers, where the libraries(ZEngine and Tetragrama) expose their Interface(the -I flag) as the root of the package and header naming is synchronized with namespaces
- This is especially important in libraries like Tetragrama where you have headers from ZEngine and Tetragrama in the same header file which is imported from a third-party project like Obelisk.
- Make the CMake files cleaner and more consistent by having a uniform naming convention for the Targets
- Remove pre-compiled headers in order to provide a good foundation for C++ Modules instead.

This commit touches on many files, but the work done is mainly adding the name of the project and the full path to the actual header file. e.g. going from `#include <ZEngineDef.h>` to `#include <ZEngine/ZEngineDef.h>`

Related to #310 , there may be need to harmonize the structure of the CMakeFiles, either all match the ZEngine format with the Main subdirectory defining the project and then containing another subdirectory with the targets or to use the format in Tetragrama and Obelisk where the targets and project are in the main Subdirectory.


.github/workflows/Engine-CI.yml
```diff
@@ -2,7 +2,7 @@ name: Engine CI
 
 on:
   push:
-    branches: [ master, develop ]
+    branches: [ master, develop, header-reorganization-mine ]
```
JeanPhilippeKernel:
```suggestion
    branches: [ master, develop ]
```

Obelisk/CMakeLists.txt
JeanPhilippeKernel: since we are using `FILE_SET HEADERS BASE_DIRS` we want to update the min version ?  it has been added since 3.23

JeanPhilippeKernel (changes requested):

Tetragrama/CMakeLists.txt
JeanPhilippeKernel: same comment here for the min version

JeanPhilippeKernel (approved):
Thanks for the re-org !

PR #504, opened 2026-02-08
ShaderC dependencies and Shader Compilation Refactor

### LINKED ISSUE #481: Duplicate Production of Shader Compilation Tools
The current shader workflow involves downloading pre-compile tools from ShaderC repositories. We are also compiling the same in the generation of dependencies(GLSLang, SPIRV..etc). This duplication costs alot of disk space and/or compilation time.

It may make sense to do away with either the download of pre-compiled binaries or do away with the compilation of the tools as dependencies.

Compiling our own will mean tighter version control across platforms reducing variables and flakiness due to different platforms.

Use of the precompiled versions requires careful selection of versions across platforms but reduces the build times significantly especially in CI builds and/or developer builds

MathewBensonCode:
- We remove the use of the shaderc download, which was only being used for compiling shaders.
 - We instead use the glslang compiler that we we're compiling anyway.
 - This way we can extract the glslang compiler that we have compiled and do the shader compilation as part of the CMake Scripting.
 - This should reduce the compilation times and/or storage requirements.
 - As Currently configured the Shaders will be compiled in each build, however, we can improve on this and ensure that compilation is only done as required at a later stage. 

Should Close #481


MathewBensonCode:
I initially attempted to use the ShaderC library to reduce the compilation by trying to use the Iibraries in the ShaderC package but ran into various linking errors on the different systems. I had worked locally for me though.

It was also not possible to use it with the macOS arm build as the ShaderC libraries available online are only x86-64.

So I went with the option of removing the ShaderC download step and shader compilation using powershell and instead incorporate it into the CMake workflow

ZEngine/ZEngine/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
 file (GLOB_RECURSE HEADER_FILES_LIST CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/*.h)
 file (GLOB_RECURSE CPP_FILES_LIST CONFIGURE_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/*.cpp)
-file (GLOB_RECURSE RESOURCE_FILES_LIST CONFIGURE_DEPENDS ${PROJECT_SOURCE_DIR}/../Resources/Shaders/*.*)
 
 source_group (TREE ${PROJECT_SOURCE_DIR}/ZEngine PREFIX "Source Files" FILES ${HEADER_FILES_LIST} ${CPP_FILES_LIST})
-source_group (TREE ${PROJECT_SOURCE_DIR}/../Resources PREFIX "Resources Files" FILES ${RESOURCE_FILES_LIST})
```
JeanPhilippeKernel, line 6: we want this - so Visual Studio will create the filter tree

Resources/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+file (GLOB VertexShaders CONFIGURE_DEPENDS "${ShaderPath}/*.vert")
+file (GLOB FragmentShaders CONFIGURE_DEPENDS "${ShaderPath}/*.frag")
+
+foreach(VertexShader ${VertexShaders})
+
+      cmake_path(GET VertexShader STEM shaderfilename)
+      cmake_path(APPEND_STRING shaderfilename "_vertex.spv" OUTPUT_VARIABLE VertexShaderOutputFileName )
+      cmake_path(APPEND ShaderOutputPath ${VertexShaderOutputFileName} OUTPUT_VARIABLE VertexShaderOutputPath)
+
+      add_custom_target("Compile-${shaderfilename}-vertex" ALL COMMAND glslang-standalone "-I${ShaderPath}" "-V" "-o" "${VertexShaderOutputPath}" "${VertexShader}")
```
JeanPhilippeKernel: Like in the ShaderCompile script, we want to drive this part with Build Config Type ,

Resources/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+endforeach()
+
+foreach(FragmentShader ${FragmentShaders})
+
+    cmake_path(GET FragmentShader STEM shaderfilename)
+    cmake_path(APPEND_STRING shaderfilename "_fragment.spv" OUTPUT_VARIABLE FragmentShaderOutputFileName )
+    cmake_path(APPEND ShaderOutputPath ${FragmentShaderOutputFileName} OUTPUT_VARIABLE FragmentShaderOutputPath)
+
+
+    add_custom_target("Compile-${shaderfilename}-fragment" ALL COMMAND glslang-standalone "-V" "-I${ShaderPath}" "-o" "${FragmentShaderOutputPath}" "${FragmentShader}")
```
JeanPhilippeKernel, line 34: Same comment here as above

JeanPhilippeKernel (changes requested):

JeanPhilippeKernel (approved):
Thanks a lot for this awesome PR 🎉

PR #506, opened 2026-02-16
refactored Command Buffer logic

### LINKED ISSUE #172: Make the rendering pipeline multi-threaded
Right now,
the rendering pipeline is single-threaded, we can improve it and make it faster by using more threads.

### LINKED ISSUE #500: CommandBufferManager logic is buggy
The CommandBufferManager has several design and implementation issues that cause or crashes in a of the  Vulkan renderer.

**Race Conditions in Instant Commands**
GetInstantCommandBuffer uses a condition variable to wait for m_instant_semaphore->GetState() == Idle, but lacks proper multi-producer protection. Multiple threads calling it simultaneously could:
- All pass the predicate check before any submission completes.
- Overwrite m_executing_instant_command flag inconsistently.
- Submit overlapping buffers to the same semaphore/fence, corrupting GPU synchronization.

JeanPhilippeKernel:
In this PR, we reviewing the whole logic behind the Command Manager that helps to create and distribute Command Buffer across engine Main Thread, Render Thread and Worker threads.

This new approach give us more flexibilities and make things for explicit about which, how and how many CB run per thread given an InFlight frame.

In addition to the Command Buffer Manager fixes, we had to review the Device swapchain creation and how it reacts to the Window resizing request - 

all the mentioned above open the road to introduce the multi-threading support to engine - 
This PR introduces support of dedicated Render Thread which is focusing on processing RenderPayload data and execute GPU commands.


JeanPhilippeKernel:
Hey @MathewBensonCode 
it'd be great if you can help test this branch on your AMD drivers  Linux/Windows -  been seeing drivers crashing on Intel.
@jnyfah  is already reporting flickering on [Linux -Intel](https://github.com/JeanPhilippeKernel/RendererEngine/issues/507) :

MathewBensonCode:
Hi, I will have a look and let you know by tomorrow.


> Hey @MathewBensonCode it'd be great if you can help test this branch on your AMD drivers Linux/Windows - been seeing drivers crashing on Intel. @jnyfah is already reporting flickering on [Linux -Intel](https://github.com/JeanPhilippeKernel/RendererEngine/issues/507) :

Sorry for taking some time to revert. I was having a problem with my link to github so when I was trying to run the workflow, the connection kept dropping. 

Today morning I used the build artifacts from the CI build. The Release version crashed for some reason, but the Debug one worked fine on linux and AMD. 

So I haven't experienced the flickering issue.

However, for linux flickering issue, it may be possible that the wrong driver is being used, i.e. the system one, which can be newer.

The library as we build it ships with a specific version that maps to the other libraries. So when using the packaged libraries we have to set the LD_LIBRARY_PATH environment variable with the location of the libraries. If you observe the CI workflow for linux we set this in order to tell the system where to find the correct driver.

Also checking which libraries are being used with the binary using the command `ldd Obelisk` is useful for debugging these kind of issues.

jnyfah:
@JeanPhilippeKernel  @MathewBensonCode 

tried setting the environment variable and it really improved the graphics a lot but still a bit distorted 

image

MathewBensonCode:
It looks like the issue is with the actual system  display driver.

The "vulkan driver" that we install and link to in the application is really just a loader whose job is to find the actual driver in the system. The actual driver in the system is usually called mesa*

The way our application is currently set up is to use glfw 3.3.10 which uses the x11 mode by default. Alot of linux desktop environments now use wayland, which means that x11 applications rely on the `xwayland` implementation of the desktop environment(gnome or kde) that you have installed. Perhaps try update the packages in the system and try different ui environments like i3 window manager, xfce and/or sway.

Is the problem only happening with this particular commit?

JeanPhilippeKernel:
Thanks @MathewBensonCode and @jnyfah  for testing - 

Actually, the issue wasn't related to  the driver, but mostly how we allocate texture resources-
at the origin, we were using Combined texture sampler, which is great when we want to combine the texture + sampler directly. but the issue with it is this approach doesnt scale enough for a engine - let's say we want  to handle 600 textures - it will have the consequence to create 600 textures +  600 samplers !! 

On some drivers, like intel, it crashes immediately, while others will weird things like on @jnyfah  screenshot - and some will simply silence the issue - macOS.

In addition the issue above, there were some threading issues between the RenderThread and MainThread and ownership of vulkan resources - 
Most of it have been solved by redesigning the `AsynchronousLoader class` to ensure that each resource is available per frame -per thread - per pool

For better coordination, I had to introduce the use of Timeline semaphore and Binary semaphore for the rendering and presentation operation.

While this new way of rendering is great and stable than the previous (again Intel driver is very strict about usage of timeline semaphore, a small mistake or omitted config, it crashes everything), it surfaces another threading issue around the `HandleManager` which helps a lot for the texture/attachement resizing for the RenderGraph

Will address the fix for in a seperate PR because  it involves solving a ABA problem #514

jnyfah:
@JeanPhilippeKernel  do you want to merge this before fixing the HandleManager / ABA issue??

JeanPhilippeKernel:
@jnyfah , yes
This PR is already enough and complex , 

Will handle the ABA stuff separately for clarity

jnyfah (approved):

PR #509, opened 2026-03-24
Make Builds Run in Parallel

MathewBensonCode:
- Use CMake's `--parallel` flag to allow MSVC and/or Apple based builds to work in parallel. 
- Ninja on Linux works in parallel by default and that is why its CI works alot faster.

### DIFF
--- a/Scripts/BuildEngine.ps1
+++ b/Scripts/BuildEngine.ps1
@@ -140,7 +140,9 @@ function Build([string]$configuration, [int]$VsVersion , [bool]$runBuild) {
     
         Write-Host "Building $systemName $architecture $configuration"
 
-        $buildArguments = "--build --preset $configName"
+        $processorCount = [ENVIRONMENT]::ProcessorCount
+
+        $buildArguments = "--build --preset $configName --parallel $processorCount"
 
         $buildProcess = Start-Process $cMakeProgram -ArgumentList $buildArguments -NoNewWindow -PassThru
 


JeanPhilippeKernel:
Should we not just extend support of ninja to Windows and Apple build? 

Thanks for PR because it drags my attention to this (.... focusing on fixing the CommandManager threading issues)  build time once again.
I was planning to create an issue to add support of ninja for both Windows and Apple builds

MathewBensonCode:
I think ideally, now that we have moved a lot of the functionality into CMake, it may perhaps be better to try and make it so that we rely in the powershell even less. This way the user should be able to pick their generator of choice and just use CMake as they desire.

I had experimented with some CMake Presets using Ninja on windows so I could possibly add that for now. The CMake Preset method is actually very crucial for those who use VSCode. 

Now that we are having newer and more up-to-date dependencies, we may need to reduce the CMake Options we are specifying in the CMake Preset File.

I actually use ninja on windows as well, but my reason is generally because of my workflow which needs the generation of the `compile_commands.json` file which is generated by the Ninja. I usually bypass the powershell file and just do the cmake generate and build commands. I don't think I pass any arguments and it actually builds fine.

MathewBensonCode:
So, I have been trying to see how to bring in Ninja to both Apple and Windows and it may need too much work with not so clear gains over the solution in this pull request. 

In other words, even if we we're to pursue this, it will require a whole new pull-request as we may need to change the powershell script parameters, create new presets in the cmake-preset file, test and modify the CI builds, etc. 

So for now I think it is better to merge this solution and have each system have its most native build system, but accelerated using the `--parallel` flag to handle all the internal complexities.

JeanPhilippeKernel:
alright, 
but let's create a task to track the ninja support investigation for new comers -

JeanPhilippeKernel (approved):

PR #575, opened 2026-08-06
feat(engine): move VFS into EngineContext, pass MemoryManager to Engine::Initialize

JeanPhilippeKernel:
## Summary

- Moves `VFS` (and its arena) from `GameApplication` into `EngineContext`; access via `Engine::GetContext()->VFS`
- Changes `Engine::Initialize` to accept `MemoryManager*` instead of `ArenaAllocator*`
- Adds `Engine::GetContext()` static accessor; removes `GameApplication::GetVFSContext()`
- Adds Step 6 precondition asserts: `memory != nullptr`, `Logger::IsInitialized()`, `ThreadPoolHelper::IsInitialized()`
- Corrects init order: GameWindow (9) → VulkanDevice (10) → VFS (11) → AssetManager (12) → AppRenderPipeline (22)
- Stores VFS arena in `EngineContext::VFSArena` to keep lifetime correct (was a dangling local)
- Updates `Tetragrama/Editor.cpp` and `ProjectViewUIComponent.cpp` to use `Engine::GetContext()->VFS`

Depends on: #574

## Test plan

- [x] Editor boots and mounts VFS without crash
- [x] `Engine::GetContext()->VFS` is non-null after `Engine::Initialize`
- [x] Assert fires if `Engine::Initialize` is called without prior `Logger::Initialize`
- [x] Assert fires if `Engine::Initialize` is called without prior `ThreadPoolHelper::Initialize`

### DIFF
--- a/ZEngine/ZEngine/Applications/GameApplication.h
+++ b/ZEngine/ZEngine/Applications/GameApplication.h
@@ -4,7 +4,6 @@
 #include <ZEngine/Core/Memory/Allocator.h>
 #include <ZEngine/Core/Memory/MemoryManager.h>
 #include <ZEngine/Core/TimeStep.h>
-#include <ZEngine/Core/VFS/VFSContext.h>
 #include <ZEngine/Rendering/Scenes/GraphicScene.h>
 #include <ZEngine/Windows/CoreWindow.h>
 #include <ZEngine/Windows/WindowConfiguration.h>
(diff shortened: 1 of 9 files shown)


ZEngine/ZEngine/Applications/GameApplication.h
```diff
@@ -40,9 +39,6 @@ namespace ZEngine::Applications
         Controllers::ICameraControllerPtr CameraController    = nullptr;
         Rendering::Scenes::RenderScenePtr CurrentScene        = nullptr;
 
-        Core::VFS::VFSContext             VFS                 = {};
-        Core::VFS::IVFSContext*           GetVFSContext();
```
JeanPhilippeKernel, line 44: @jnyfah I've moved this to the engine -  you want to rebase some of your current WIP

Issue #649, opened 2026-08-22
feat(ecs): implement ComponentReflectionRegistry


JeanPhilippeKernel:
## Dependency position

```
#648 (FieldType / ComponentMeta types)
  └──► THIS ISSUE ──► #650 (Register built-in components)
                   └──► #651 (Inspector migration)
```

**Start after #648 is merged.** Can be worked immediately after #648 since that PR is just headers.

---

## Why

The registry is the central store all component metadata flows through. The inspector calls `ForEach` on it. Registration code in component `.cpp` files calls `Register` on it. Without it, neither #650 nor #651 can proceed.

### `ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.h`

```cpp
#pragma once
#include <ZEngine/ECS/Reflection/ComponentMeta.h>
#include <ZEngine/ECS/ComponentTypeID.h>
#include <ZEngine/Core/Containers/Array.h>
#include <ZEngine/Core/Memory/Allocator.h>

namespace ZEngine::ECS {

    class ComponentReflectionRegistry {
    public:
        static ComponentReflectionRegistry& Get();

        void Initialize(Core::Memory::ArenaAllocator* arena);

        // Register metadata. meta.Fields must outlive this registry
        // (use static arrays at the call site).
        // Second registration with the same TypeID is a no-op (debug warning logged).
        void Register(const ComponentMeta& meta);

        [[nodiscard]] const ComponentMeta* Lookup(ComponentTypeID id) const;
        [[nodiscard]] const ComponentMeta* LookupByName(const char* type_name) const;

        template<typename Fn>
        void ForEach(Fn&& fn) const {
            for (uint32_t i = 0; i < m_metas.size(); ++i)
                fn(m_metas[i]);
        }

        [[nodiscard]] uint32_t Count() const;

    private:
        Core::Containers::Array<ComponentMeta> m_metas;
        Core::Memory::ArenaAllocator*          m_arena = nullptr;
    };

}
```

### `ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp`

- `Get()` returns a static instance (initialized once).
- `Initialize(arena)` stores the arena and calls `m_metas.init(arena, 64)`.
- `Register` checks for duplicate TypeID (debug assert/warn), then `m_metas.push(meta)`.
- `Lookup` iterates `m_metas` linearly (N ≤ 64, cache-friendly, no hash needed).
[...]

PR #662, opened 2026-08-24
feat(ecs): ComponentReflectionRegistry

jnyfah:
#649

### DIFF
--- a/ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp
+++ b/ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp
@@ -7,6 +7,8 @@ namespace ZEngine::ECS
 {
     ComponentReflectionRegistry& ComponentReflectionRegistry::Get()
     {
+        // Do not access after Engine::Deinitialize() — s_instance has static storage
+        // duration and outlives the arena, leaving m_metas.m_data dangling.
         static ComponentReflectionRegistry s_instance;
         return s_instance;
     }
--- a/ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp
+++ b/ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp
@@ -1,0 +1,69 @@
+#include <ZEngine/ECS/Reflection/ComponentReflectionRegistry.h>
+#include <ZEngine/Helpers/MemoryOperations.h>
+#include <ZEngine/Logging/LoggerDefinition.h>
+#include <ZEngine/ZEngineDef.h>
+
+namespace ZEngine::ECS
+{
+    ComponentReflectionRegistry& ComponentReflectionRegistry::Get()
+    {
+        static ComponentReflectionRegistry s_instance;
+        return s_instance;
+    }
+
+    void ComponentReflectionRegistry::Initialize(Core::Memory::ArenaAllocator* arena)
+    {
+        ZENGINE_VALIDATE_ASSERT(arena != nullptr, "ComponentReflectionRegistry::Initialize: arena must not be null")
+        m_arena = arena;
+        m_meta.init(arena, 64);
+    }
+
+    void ComponentReflectionRegistry::Register(const ComponentMeta& meta)
+    {
+        ZENGINE_VALIDATE_ASSERT(m_arena != nullptr, "ComponentReflectionRegistry::Register: Initialize() must be called first")
+        ZENGINE_VALIDATE_ASSERT(meta.TypeName != nullptr, "ComponentReflectionRegistry::Register: meta.TypeName must not be null")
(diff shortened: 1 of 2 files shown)


JeanPhilippeKernel (commented):
## Review

Ready to merge with one note that needs addressing.

---

### Must address

**`Lookup` returns a pointer that can be invalidated by a subsequent `Register`**

`m_meta` is an arena-backed `Array<ComponentMeta>`. When `push` triggers `grow()`, the old data block is abandoned and a new one allocated — any `const ComponentMeta*` from a previous `Lookup` becomes dangling.

Current mitigation: `m_meta.init(arena, 64)` pre-allocates 64 slots, so grow never fires as long as fewer than 64 types are registered. With 8 built-in components this is safe, but it is a silent assumption.

**Suggested fix** — add a comment to the header making the contract explicit:

```cpp
// Pointers returned by Lookup/LookupByName remain valid for the lifetime
// of the registry, provided Register() is not called after the first Lookup
// and fewer than 64 types are registered (the pre-allocated capacity).
```

Or: accept `max_types` in `Initialize` instead of hard-coding 64.

---

### Minor (non-blocking)

**`Initialize` call site not documented** — the header gives no hint of where in the engine lifecycle `Initialize` should be called. The assert in `Register` catches a missed call at runtime, but a short comment noting the expected call site (e.g. `Engine::Initialize` alongside `ActorManager`) would help future contributors.

**Static singleton outlives the arena** — `s_instance` has static storage duration and outlives the engine arena. After `Engine::Deinitialize()`, `m_meta.m_data` is a dangling pointer. Benign today (no destructors touch it), but fragile. A comment in the `.cpp` noting this would guard against future surprises.

---

### Confirmed correct

- `Register` dedup guard — warns and returns on duplicate `TypeID` ✓
- `LookupByName` null guard before `secure_strcmp` ✓
- `ForEach` template — `Fn&&` forwarding, const iteration ✓
- `Count` impl ✓
- `Initialize` asserts on null arena and `Fields/FieldCount` consistency ✓

ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.h
```diff
@@ -0,0 +1,38 @@
...
+    {
+    public:
+        static ComponentReflectionRegistry& Get();
+
+        void                                Initialize(Core::Memory::ArenaAllocator* arena);
+
+        void                                Register(const ComponentMeta& meta);
+
+        [[nodiscard]] const ComponentMeta*  Lookup(ComponentTypeID id) const;
+        [[nodiscard]] const ComponentMeta*  LookupByName(const char* type_name) const;
```
JeanPhilippeKernel, line 24: `m_meta` is arena-backed — if `Register` is called after any `Lookup` and the array grows past 64, pointers returned by earlier `Lookup` calls become dangling. Safe with the current 8 built-in components, but add a comment documenting the 64-slot invariant so future contributors don't silently break it.

ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.h
```diff
@@ -0,0 +1,38 @@
...
+#include <ZEngine/ECS/Reflection/ComponentMeta.h>
+
+namespace ZEngine::ECS
+{
+
+    class ComponentReflectionRegistry
+    {
+    public:
+        static ComponentReflectionRegistry& Get();
+
```
JeanPhilippeKernel, line 16: Add a comment noting where `Initialize` must be called in the engine lifecycle — e.g. "Called from `Engine::Initialize` alongside `Scene` and `ActorManager`.". The assert in `Register` will catch a missed call at runtime, but the expected setup is not obvious from the header alone.

ZEngine/ZEngine/ECS/Reflection/ComponentReflectionRegistry.cpp
```diff
@@ -0,0 +1,69 @@
+#include <ZEngine/ECS/Reflection/ComponentReflectionRegistry.h>
+#include <ZEngine/Helpers/MemoryOperations.h>
+#include <ZEngine/Logging/LoggerDefinition.h>
+#include <ZEngine/ZEngineDef.h>
+
+namespace ZEngine::ECS
+{
+    ComponentReflectionRegistry& ComponentReflectionRegistry::Get()
+    {
+        static ComponentReflectionRegistry s_instance;
```
JeanPhilippeKernel, line 12: `s_instance` has static storage duration and outlives the engine arena. After `Engine::Deinitialize()`, `m_meta.m_data` is a dangling pointer. Benign today (no destructors touch it), but worth a one-line comment: `// Do not access after Engine::Deinitialize() — m_meta.m_data outlives the arena.`

JeanPhilippeKernel (approved):
All three items from the previous review are addressed. One non-blocking suggestion: add a `ZENGINE_VALIDATE_ASSERT(m_meta.size() < 64, ...)` inside `Register()` to enforce the documented 64-slot capacity constraint at runtime rather than relying solely on the comment. Fine to do as a follow-up. Ready to merge.

PR #747, opened 2026-09-04
fix(logging): no-op Logger::Log when not initialized

### LINKED ISSUE #746: Logger::Log crashes via infinite recursion when called before Logger::Initialize
## Summary

`Logger::Log` (`ZEngine/ZEngine/Logging/Logger.cpp:161-193`) has no guard against being called before `Logger::Initialize()` has run (or after `Logger::Dispose()`). Found while adding a new GTest suite (`GpuAllocatorTest`, part of this month's rendering-foundation work) that triggered a `ZENGINE_CORE_WARN` call in a bare test binary — a prior test suite's `TearDown()` had already called `Logger::Dispose()`, leaving the global ring buffers (`s_log_message_rb`, `s_log_raw_string_rb`) at their default-constructed empty state (`size() == 0`).

## Root cause

`Logger::Log:180`:
```cpp
auto buffer_index = s_ring_buffer_index.value.fetch_add(1, std::memory_order_relaxed) % s_log_message_rb.size();
```
If `s_log_message_rb.size() == 0` (uninitialized), this is a mod-by-zero. Whatever value results, the subsequent `s_log_raw_string_rb[buffer_index]` / `s_log_message_rb[buffer_index]` indexing (`Logger.cpp:181-184`) is then out-of-bounds against a zero-capacity array, which trips `Array<T>::operator[]`'s own `ZENGINE_VALIDATE_ASSERT(index < m_size, "Index out of range")` bounds check. That assert's own error-formatting path calls back into `Logger::Log` to report the failure — which hits the exact same zero-sized array again — recursing until the stack overflows (`SIGSEGV`, "Could not determine thread index for stack guard region").

## Repro

1. In a GTest binary, run any test suite that calls `Logger::Initialize()` then `Logger::Dispose()` in its fixture (e.g. `ZEngine/tests/Logging/Logger_test.cpp`).
2. In a *later* test suite (no `Logger::Initialize()` call of its own), trigger any `ZENGINE_CORE_WARN`/`ZENGINE_CORE_ERROR`/etc. call.
3. Process crashes with a stack-overflow `SIGSEGV`, with a repeating `Logger::Log` ↔ `Array<String>::operator[]` call stack.

## Suggested fix

Guard the top of `Logger::Log` with `if (s_log_message_rb.size() == 0) return;` (or check `Logger::IsInitialized()`, which already exists), so a pre-init/post-dispose log call is a silent no-op instead of a crash. Test-suite ordering shouldn't be able to crash unrelated tests just because one of them exercises the Logger lifecycle.

## Where

`ZEngine/ZEngine/Logging/Logger.cpp:161-193` (`Logger::Log`)

BetterAndBetterII:
## Summary

`Logger::Log` indexed the ring buffer with `fetch_add(...) % s_log_message_rb.size()` without checking initialization. Before `Initialize()` the capacity is zero (modulo-by-zero / `SIGFPE`). After `Dispose()` the same path can recurse through assert macros back into `Log` until the stack overflows.

Return immediately when `!IsInitialized()`. `IsInitialized()` already exists; this just uses it at the top of `Log`.

## Test plan

- [x] `LoggerUninitialized.LogBeforeInitializeDoesNotCrash` — was RED (`SIGFPE`), now GREEN
- [x] `LoggerTest.LogAfterDisposeDoesNotCrash` — GREEN
- [x] Existing `LoggerTest.*` still pass (15/15 in the isolated Logger binary)

```bash
ZEngineTests --gtest_filter='LoggerUninitialized.*:LoggerTest.*'
```

Fixes #746

### DIFF
--- a/ZEngine/ZEngine/Logging/Logger.cpp
+++ b/ZEngine/ZEngine/Logging/Logger.cpp
@@ -160,6 +160,11 @@ namespace ZEngine::Logging
 
     void Logger::Log(LogChannel channel, LogLevel level, std::string_view msg)
     {
+        if (!IsInitialized())
+        {
+            return;
+        }
+
         const int min = k_min_level[static_cast<size_t>(channel)];
         if (static_cast<int>(level) < min)
         {
--- a/ZEngine/tests/Logging/Logger_test.cpp
+++ b/ZEngine/tests/Logging/Logger_test.cpp
@@ -51,6 +51,18 @@ protected:
 std::string LoggerTest::s_log_dir;
 std::string LoggerTest::s_crash_dir;
 
+TEST(LoggerUninitialized, LogBeforeInitializeDoesNotCrash)
+{
+    if (Logger::IsInitialized())
+    {
+        Logger::Dispose();
+    }
+
+    EXPECT_FALSE(Logger::IsInitialized());
+    Logger::Log(LogChannel::ENGINE, LogLevel::WARN, "before initialize");
+    ZENGINE_CORE_WARN("before initialize via macro");
+}
+
 TEST_F(LoggerTest, IsInitializedAfterInit)
 {
     EXPECT_TRUE(Logger::IsInitialized());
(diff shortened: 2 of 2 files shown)


JeanPhilippeKernel (approved):
Thanks for the fix @BetterAndBetterII !!

Issue #719, opened 2026-09-01
test(scene): serialization validation and recovery matrix


JeanPhilippeKernel:
## Summary

Create the production serialization validation matrix. The existing fixed
eight-test sketch is insufficient because durable identity, atomic persistence,
transactional reconstruction, compatibility, cook, and render-derived state all
need independent negative and integration coverage.

## Scope

- Fresh-process YAML round-trip for scene UUID, entity UUIDs, components,
  hierarchy semantics, asset UUIDs, and `SceneGridSettings`; verify that runtime
  IDs/handles are not persisted.
- Canonical-save determinism, interrupted-write recovery, and dirty-checkpoint
  behavior.
- Negative validation: duplicate/malformed UUIDs, invalid references, cycles,
  non-finite values, illegal enums, oversized input, malformed component data,
  and truncated binaries.
- Compatibility: unknown-data preservation or explicit save rejection, plus
  deterministic migration fixtures.
- Transactional load/reload/Play Stop: failed load preserves the live scene;
  successful load rebuilds hierarchy, actor/editor proxies, render bindings, and
  invalidates stale epoch-bound targets.
- Cook: unresolved assets and unsupported components fail; valid artifacts load
  and preserve durable semantics.
- Registry/codec unit coverage for explicit startup registration, stable schema
  ordering, built-in codecs, and UUID reference hooks.

## Dependencies

- #714, #715, #716, #717, #718, #829, and #830

## Acceptance criteria

- Tests run in the normal engine test target and include fresh-process coverage
  where process-local component IDs could otherwise mask a defect.
- Every documented validation failure returns a structured diagnostic and never
  partially publishes a scene.
- The suite covers the source and cooked formats independently.

## Reference

`ZEngine/docs/future-plan/scene-serialization.md` section 11.

Issue #718, opened 2026-09-01
feat(scene): scene cook validation and package manifest


JeanPhilippeKernel:
## Summary

Implement the scene cook boundary: consume validated YAML source, validate the
complete asset/reference graph, emit the stable binary artifact, and record the
package manifest inputs needed for correct incremental cooking.

## Scope

- Consume the canonical source model and validation produced by #715 and #830.
- Resolve asset UUIDs through the registry; editor placeholders are never
  acceptable in a shipping cook.
- Emit the binary artifact through #716 only after schema/reference validation
  succeeds.
- Write package-manifest records with source identity/content fingerprint plus
  relevant importer/schema/build inputs, so incremental cook cannot reuse a
  stale artifact.
- Write artifacts and manifest updates atomically; leave prior valid artifacts
  intact on failure.
- Report source path, entity/component/schema key, and asset UUID in cook
  diagnostics without treating a local native path as scene data.

## Non-goals

- Do not make `YAMLSceneSerializer` editor-only if command-line/tool cooking
  needs source parsing; keep editor UI separate from authoring-format support.
- Do not add ad hoc binary layout or duplicate schema validation in the cooker.

## Dependencies

- #715 — versioned YAML SceneDocument and canonical save
- #716 — stable-schema binary artifact
- #830 — compatibility/migration policy

## Acceptance criteria

- A valid source scene produces a deterministic runtime artifact and manifest.
- Missing/invalid/unsupported assets or components fail cook with an actionable
  error and no partially published artifact.
- Incremental cook invalidates when the source, relevant schema, or relevant
  import settings change.

## References

`ZEngine/docs/future-plan/scene-serialization.md` sections 5, 8, 9, and 11.

Issue #717, opened 2026-09-01
feat(scene): built-in component codecs and UUID reference hooks


JeanPhilippeKernel:
## Summary

Register production YAML/binary codecs and UUID-reference hooks for every
currently authored built-in ECS component. The existing reflection registrations
are not serializer registrations, and static initialization is not an accepted
production lifecycle.

## Scope

- Add explicit built-in scene-schema registration after #714 has been initialized
  during engine startup.
- Implement codecs for the current built-ins: `TransformComponent`,
  `MeshComponent`, `CameraComponent`, `LightComponent`, `MaterialComponent`,
  `NameComponent`, `RigidBodyComponent`, and `UUIDComponent`.
- Classify authored, runtime-derived, editor-only, and forbidden fields. Runtime
  fields such as `RenderInstanceId` and GPU handles are excluded.
- Validate finite transforms, legal component values, UUID syntax, asset UUID
  references, and component-specific constraints.
- Implement stable field keys and UUID reference enumerate/patch hooks consumed
  by staged load, duplication, undo/redo, and Play restoration.
- Treat UUID component data as durable identity, enforce uniqueness through the
  scene-document path, and keep `EntityID` only in loader/runtime maps.

## Non-goals

- Do not use static C++ registration, `std::function`, reflection offsets, or
  raw-byte serialization.
- Do not introduce a separate parent/hierarchy format outside the scene document
  and staged resolver.

## Dependencies

- #714 — stable component schema registry

## Acceptance criteria

- All built-in authored values round-trip through stable YAML and binary codecs.
- Runtime-derived/render handles are absent after serialization and rebuilt by
  #829.
- Invalid component payloads produce structured errors rather than being dropped.
- Tests cover codec values, field keys, reference enumeration, and rejection
  cases for each built-in component.

## References

`ZEngine/docs/future-plan/scene-serialization.md` sections 2, 4, 6, and 11.

Issue #716, opened 2026-09-01
feat(scene): stable-schema binary scene artifact


JeanPhilippeKernel:
## Summary

Implement the validated stable-schema binary scene artifact used by the runtime
cook path. It is a portable cooked representation of the authoring document,
not a raw-memory dump and not a zero-parse promise.

## Scope

- Define a binary header with magic, version, endian/feature compatibility, and
  integrity checks.
- Encode bounded offsets/lengths, stable component schema keys and versions,
  scene UUID/settings, and UUID mappings required for references.
- Validate every header, table, count, offset, length, version, and component
  payload before constructing runtime data.
- Reuse the component codecs/schema registry from #714 and #717; never encode
  runtime `ComponentTypeID`, `EntityID`, pointers, alignment-dependent structs,
  or raw component bytes as the durable contract.
- Honor the compatibility/migration policy from #830 before candidate scene
  reconstruction or cook output.

## Non-goals

- Do not define a separate authoring schema; YAML source is owned by #715.
- Do not make unresolved assets an assertion-driven runtime path; cook reports
  a structured validation failure.
- Do not wire package staging/manifest ownership here; #718 owns that boundary.

## Dependencies

- #714 — stable component schema registry
- #717 — built-in component codecs and UUID reference hooks
- #830 — unknown-data preservation and schema migrations

## Acceptance criteria

- Valid cooked data loads on a fresh process with preserved durable identity and
  component semantics.
- Truncated, oversized, malformed, unsupported-version, and incompatible data
  fails safely without partial publication.
- Binary assets remain independent of process-local type allocation and native
  ABI/layout differences.

## References

`ZEngine/docs/future-plan/scene-serialization.md` sections 8, 9, and 11.

PR #765, opened 2026-09-06
feat(scene): SceneSnapshot

### LINKED ISSUE #713: feat(scene): SceneSnapshot struct + ISceneSerializer interface — foundation
## Summary
Create the two foundation types for the new scene serialization system. Neither exists today.

## SceneSnapshot
```cpp
// ZEngine/ZEngine/ECS/SceneSnapshot.h
struct SceneSnapshot {
    uuids::uuid                         SceneUUID = {};
    Core::Containers::String            Name      = {};
    Core::Containers::Array<EntityID>   Entities  = {};
};
```
Lightweight snapshot that the serializer works with. Separate from `ZEngine::ECS::Scene` (the live runtime manager). The serializer snapshots `Scene → SceneSnapshot` on save and reconstructs `Scene` entities on load.

## ISceneSerializer
```cpp
// ZEngine/ZEngine/ECS/ISceneSerializer.h
class ISceneSerializer {
public:
    virtual Core::VFS::VFSResult<void> Serialize(
        Core::VFS::IVFSContext&, const Core::VFS::VFSPath&, const SceneSnapshot&) = 0;
    virtual Core::VFS::VFSResult<void> Deserialize(
        Core::VFS::IVFSContext&, const Core::VFS::VFSPath&, SceneSnapshot&) = 0;
};
```

## References
scene-serialization.md §1 and §2

jnyfah:
## Summary

Implements the foundation of the scene serialization system. Closes #713, #714, #715. Partial coverage of #719 (YAML tests only — binary tests pending #716).

- **`SceneSnapshot`** (`ECS/SceneSnapshot.h`) — lightweight snapshot struct the serializer works with, separate from the live `Scene` runtime manager.
- **`ISceneSerializer`** (`ECS/ISceneSerializer.h`) — abstract interface with `Serialize` / `Deserialize` over VFS paths.
- **`ComponentSerializerRegistry`** (`ECS/ComponentSerializerRegistry.h/.cpp`) — dynamic dispatch table mapping `ComponentTypeID` to C-style YAML and binary callbacks. No `std::function`. Arena-backed. Requires `Initialize(arena)` before registration.
- **`YAMLSceneSerializer`** (`ECS/YAMLSceneSerializer.h/.cpp`) — editor-only (`#ifdef ZENGINE_EDITOR`) YAML serializer using `yaml-cpp`. Validates all `*_uuid` fields before mutating the scene. Unknown component keys are skipped for forward compatibility. Asset references are warned on but not blocked when not registered.
- **CMake:** `ZENGINE_BUILD_EDITOR` option → `ZENGINE_EDITOR=1` define gates editor-only code.
- **Tests:** `Scene/ComponentSerializerRegistryTest.cpp` (8 tests — registry lookup, dedup, ForEach, YAML + binary round-trips, hand-edited YAML, emitted YAML) and `Scene/YAMLSceneSerializerTest.cpp` (8 tests — serialize/deserialize round-trip, hand-edited YAML, bad asset ref rejection, unknown component skip, malformed YAML, missing file).

## Out of scope (tracked separately)
- #716 — `BinarySceneSerializer`
- #717 — register YAML + binary fns for the 8 built-in components
- #718 — scene cook pipeline
- Binary half of #719 tests

## Test plan
- CI build passes on Linux / Windows / macOS with `ZENGINE_BUILD_EDITOR=ON`
- All 16 new tests pass
- Existing 589 tests unaffected


JeanPhilippeKernel (changes requested):
## Review — 3 confirmed bugs, 3 items to address

### Bug 1 (confirmed) — `ComponentSerializerRegistryTest.cpp:633` + `YAMLSceneSerializerTest.cpp:953`
**Double-Initialize resets the singleton registry across test files.**

Both files call `ComponentSerializerRegistry::Get().Initialize()` with separate static `MemoryManager` instances. The two one-time guards are independent — whichever fires second resets `m_ids`/`m_fns` to a different arena, discarding all prior registrations. Test order determines which init wins.

**Fix:** extract a single shared init helper used by both fixtures, or add an `IsInitialized()` guard inside `Initialize()` so subsequent calls are no-ops.

---

### Bug 2 (confirmed) — `SceneSnapshot.h:12` (`Entities` field)
**`SceneSnapshot::Entities` crashes on `push` without explicit `init`.**

`Array<EntityID> Entities = {}` has `m_data=nullptr, m_capacity=0` until `init(arena, N)` is called. Any caller writing `snap.Entities.push(id)` directly asserts in Debug or corrupts in Release.

**Fix:** add a doc note or a `Create(arena, capacity)` factory:
```cpp
/// @note Call Entities.init(arena, N) before push.
Core::Containers::Array<EntityID> Entities = {};
```

---

### Bug 3 (confirmed) — `YAMLSceneSerializer.cpp:214`
**`LooksLikePath` is dead code.**

A valid UUID contains only hex digits and hyphens — never `/`, `\`, or `.`. Any string passing `from_string` cannot trigger `LooksLikePath`; any string failing it is caught by `!parsed.has_value()` anyway.

**Fix:** remove `LooksLikePath` entirely. The condition simplifies to:
```cpp
if (!parsed.has_value())
{
    ZENGINE_CORE_ERROR("YAMLSceneSerializer: '{}' must be a UUID, found '{}'", key, value);
    return VFSResult<void>::Fail(VFSError::Corrupted);
}
WarnIfAssetMissing(parsed.value());
```

---

### Should fix — `YAMLSceneSerializer.cpp:414` (emitter)
**YAML block style not explicitly enforced.**

The test asserts no `{` but relies on yaml-cpp's default. The spec requires explicit block style.

**Fix:**
```cpp
emitter.SetIndent(2);
emitter << YAML::Block << root;
```

---

### Should fix — `YAMLSceneSerializer.cpp:519` (entity loop)
**No rollback if a component deserialization callback throws.**

`CreateEntity()` runs before component callbacks. `YAML::Node::as<float>()` throws `YAML::BadConversion` on malformed values (e.g. `position: [bad, 2, 3]`). Pre-validate checks UUIDs but not numeric values — scene ends up with orphaned entities.

**Fix:** wrap the per-entity deserialization block in `try/catch(const YAML::Exception&)`, call `m_scene->DestroyEntity(id)` and log a warning on catch.

---

### Should fix — `ComponentSerializerRegistry.cpp:36` (Initialize capacity)
**64-slot capacity limit is undocumented and untunable.**

The assert fires before `push()` would auto-grow. A caller registering a 65th component gets an opaque assert with no guidance.

**Fix:** expose capacity as a parameter:
```cpp
void Initialize(Core::Memory::ArenaAllocator* arena, uint32_t capacity = 64);
```

ZEngine/tests/Scene/ComponentSerializerRegistryTest.cpp
```diff
@@ -0,0 +1,266 @@
...
+using namespace ZEngine::Core::Memory;
+
+namespace
+{
+    ComponentSerializerRegistry& Registry()
+    {
+        static MemoryManager s_manager = [] {
+            MemoryManager m;
+            m.Initialize(ZMega(8), {});
+            ComponentSerializerRegistry::Get().Initialize(&m.MainArena);
```
JeanPhilippeKernel: **Bug (confirmed): double-Initialize resets the singleton registry.**

Both this file and YAMLSceneSerializerTest.cpp call ComponentSerializerRegistry::Get().Initialize() with separate static MemoryManager instances. The two one-time guards are independent — whichever fires second resets m_ids/m_fns to a different arena, discarding all prior registrations.

Fix: share a single init helper across both fixtures, or add an IsInitialized() guard inside Initialize() so subsequent calls are no-ops.

ZEngine/ZEngine/ECS/SceneSnapshot.h
```diff
@@ -0,0 +1,15 @@
...
+#include <ZEngine/ECS/EntityID.h>
+#include <uuid.h>
+
+namespace ZEngine::ECS
+{
+    struct SceneSnapshot
+    {
+        uuids::uuid                       SceneUUID = {};
+        Core::Containers::String          Name      = {};
+        Core::Containers::Array<EntityID> Entities  = {};
```
JeanPhilippeKernel, line 15: **Bug (confirmed): Entities crashes on push without explicit init.**

Array<EntityID> Entities = {} has m_data=nullptr, m_capacity=0 until init(arena, N) is called. Any caller writing snap.Entities.push(id) directly asserts in Debug or corrupts in Release.

Fix: add a doc comment or a Create(arena, capacity) factory so the precondition is visible at the call site.

ZEngine/ZEngine/ECS/YAMLSceneSerializer.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include <ZEngine/Managers/AssetManager.h>
+#include <ZEngine/ZEngineDef.h>
+
+using namespace ZEngine::Core::VFS;
+
+namespace ZEngine::ECS
+{
+    namespace
+    {
+        bool LooksLikePath(cstring v)
```
JeanPhilippeKernel: **Bug (confirmed): LooksLikePath is dead code.**

A valid UUID contains only hex digits and hyphens, never /, backslash, or '.'. Any string passing from_string cannot trigger this check; any string failing it is already caught by !parsed.has_value().

Fix: remove LooksLikePath entirely. The condition simplifies to just the UUID parse check + WarnIfAssetMissing on success.

ZEngine/ZEngine/ECS/YAMLSceneSerializer.cpp
```diff
@@ -0,0 +1,363 @@
...
+        UUIDToChars(scene.SceneUUID, uuid_text);
+        scene_node["uuid"]     = uuid_text;
+        scene_node["name"]     = scene.Name.empty() ? "" : scene.Name.c_str();
+        scene_node["entities"] = entities;
+
+        YAML::Node root(YAML::NodeType::Map);
+        root["scene"] = scene_node;
+
+        YAML::Emitter emitter;
+        emitter << root;
```
JeanPhilippeKernel: **Should fix: YAML block style not explicitly enforced.**

Test verifies no '{' but relies on yaml-cpp default. Add before emitting:
emitter.SetIndent(2);
emitter << YAML::Block << root;

ZEngine/ZEngine/ECS/YAMLSceneSerializer.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const auto& reflection  = ComponentReflectionRegistry::Get();
+
+        if (!entities.IsDefined() || !entities.IsSequence())
+        {
+            return VFSResult<void>::Ok(); // a scene with no entities is valid
+        }
+
+        for (const auto& entity_node : entities)
+        {
+            const EntityID id = m_scene->CreateEntity();
```
JeanPhilippeKernel, line 312: **Should fix: no rollback if a component deserialization callback throws.**

CreateEntity() runs before component callbacks. YAML::Node::as<float>() throws YAML::BadConversion on malformed values (e.g. position: [bad, 2, 3]). Pre-validate checks UUIDs but not numeric values — scene ends up with orphaned entities.

Fix: wrap per-entity deserialization in try/catch(const YAML::Exception&), call m_scene->DestroyEntity(id) and log a warning on catch.

ZEngine/ZEngine/ECS/ComponentSerializerRegistry.cpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+        ZENGINE_VALIDATE_ASSERT(m_arena != nullptr, "ComponentSerializerRegistry::Register: Initialize() must be called first")
+
+        if (Lookup(type_id))
+        {
+            ZENGINE_CORE_WARN("ComponentSerializerRegistry: TypeID {} already registered, ignoring", type_id);
+            return;
+        }
+
+        ZENGINE_VALIDATE_ASSERT(m_ids.size() < m_ids.capacity(), "ComponentSerializerRegistry::Register: exceeded reserved capacity")
```
JeanPhilippeKernel: **Should fix: 64-slot capacity limit is undocumented and untunable.**

The assert fires before push() would auto-grow. A caller registering a 65th component gets an opaque assert with no guidance.

Fix: expose capacity as a parameter: Initialize(ArenaAllocator* arena, uint32_t capacity = 64).

JeanPhilippeKernel (changes requested):
## Follow-up review — good progress, two items remaining

5 of 6 previous findings are fixed. One partial fix, one new bug introduced.

### Still needs fixing

**1. SceneSnapshot::Entities — partial fix (SceneSnapshot.h)**
A `Create()` factory was added and `Deserialize()` calls `init()` correctly. But the struct is still default-constructible without initialization — `SceneSnapshot snap{}; snap.Entities.push(id)` bypasses `Create()` and still crashes on the null internal pointer. The fix is to add a doc comment on the `Entities` field: `/// @note Use SceneSnapshot::Create() to construct — direct push() without init() asserts.`

**2. New bug: hardcoded 64-entity capacity in Deserialize**
`out_scene.Entities.init(m_arena, 64)` crashes at the `push()` assert on the 65th entity in any scene file with more than 64 entities. The entity count is available at that point via `entities.size()`:
```cpp
out_scene.Entities.init(m_arena, static_cast<uint32_t>(entities.size()) + 1);
```

### Other findings (all fixed)

- Double-Initialize → fixed via early-return guard in `Initialize()`
- LooksLikePath dead code → removed
- YAML block style → `SetIndent(2)` + `YAML::Block` applied
- Entity rollback on exception → `try/catch` + `DestroyEntity` + `Entities.pop()`
- 64-slot capacity → named parameter with default

### Note on removed YAML tests

The `YAMLSceneSerializerTest.cpp` file was removed ("remove yaml test for now"). The remaining `ComponentSerializerRegistryTest.cpp` covers the registry but not the actual serializer round-trips. The YAML tests should come back before merge, or the scope should be explicitly noted in the PR description.

JeanPhilippeKernel (approved):
All findings resolved. Last commit fixes the remaining two items: entity count initialized from the actual YAML sequence size (not hardcoded 64), inner try/catch removed from component deserialization (callbacks must use safe yaml-cpp APIs), and doc note added on SceneSnapshot::Entities. The YAML::Load exception boundary is kept as the unavoidable third-party API constraint with a comment. Ready to merge.

PR #773, opened 2026-09-08
fix(vulkan,rendering,editor): vkQueueSubmit2, deferred texture upload, async drag-drop

JeanPhilippeKernel:
## Summary

- **vkQueueSubmit2 migration** — replace \`vkQueueSubmit\` + \`VkTimelineSemaphoreSubmitInfo\` with \`vkQueueSubmit2\` + \`VkSemaphoreSubmitInfo\` in all timeline semaphore submit paths. Each semaphore carries its own struct, eliminating the parallel-array count ambiguity that caused Intel Windows driver corruption (UINT64_MAX semaphore value). Enable \`VkPhysicalDeviceSynchronization2Features\` in \`VkDeviceCreateInfo\` — the extension was loaded but the feature flag was never set. Fix type correctness: stage flag fields now use \`VkPipelineStageFlags2\`. Remove \`ASSERT_TIMELINE_MONOTONIC\` macro and all 6 call sites — root cause fixed at API level.

- **Deferred texture upload** — texture async ops and descriptor updates go into 1-frame SPSC deferred queues. \`Present()\` runs before \`SubmitAsyncUploads()\` so the same-frame GPU stall on drag-drop is eliminated. \`RequestDeferredDescriptorUpdate\` pre-fills the new descriptor slot with the fallback image view for the one-frame upload window, written from the render thread after present to avoid concurrent \`vkUpdateDescriptorSets\`.

- **Async drag-drop mesh deserialization** — \`DeserializeMeshAssetFile\` and \`IngestMaterialFromUUID\` are synchronous file reads that were blocking the render thread on every drag-drop. Both now run on a \`ThreadPool\` worker with a dedicated per-drop \`ArenaAllocator\` (4× file size + 8 MB). The main-thread callback handles only the lightweight \`IngestMesh\` + actor create.

- **Same-session texture fix** — in a same-session cook+drag-drop flow, \`GetAsset<AssetMaterial>(uuid)\` returned null because the VFSFileWatcher marks the AssetRegistry record Stale after the importer sets it Loaded. Material index lookup now uses \`UUIDToMaterialSlot.find(uuid)\` directly — the raw slot set by \`IngestMaterial\`, never touched by the scanner.

## Result

Stable 121 fps through mesh drag-drop with correct textures, both in fresh sessions and same-session cook+ingest flows.

## Test plan

- Drag a cooked \`.zemesh\` onto the viewport — FPS stays stable, mesh appears with correct textures
[...]

PR #774, opened 2026-09-09
fix(vulkan,rendering,editor): vkQueueSubmit2 migration + deferred texture upload + async drag-drop

JeanPhilippeKernel:
## Summary

- **vkQueueSubmit2 migration** — replace `vkQueueSubmit` + `VkTimelineSemaphoreSubmitInfo` with `vkQueueSubmit2` + `VkSemaphoreSubmitInfo` in all timeline semaphore submit paths (`DeviceSwapchain::Present()` and `VulkanDevice::QueueSubmit(timeline)`). Each semaphore carries its own struct, eliminating the parallel-array count ambiguity that caused Intel Windows driver corruption (UINT64_MAX semaphore value). Enable `VkPhysicalDeviceSynchronization2Features` in `VkDeviceCreateInfo` — the extension was loaded but the feature flag was never set. Fix type correctness: `AsyncGPUOperationHandle::StageFlags`, `AsyncUploadJob::WaitFlag`, and `QueueSubmit(timeline)` `wait_flag` now use `VkPipelineStageFlags2`. Remove `ASSERT_TIMELINE_MONOTONIC` macro and all 6 call sites — root cause fixed at the API level; the per-frame `vkGetSemaphoreCounterValue` calls were expensive on MoltenVK.

- **Deferred texture upload** — texture async ops and descriptor updates go into 1-frame SPSC deferred queues (`DeferredAsyncGPUOperations`, `DeferredTextureDescriptorUpdates`). `Present()` runs before `SubmitAsyncUploads()` so the same-frame GPU stall on drag-drop is eliminated. `RequestDeferredDescriptorUpdate` pre-fills the new slot with the fallback image view (pink) for the one-frame upload window so the mesh renders correctly immediately.

- **Async drag-drop mesh deserialization** — `DeserializeMeshAssetFile` and `IngestMaterialFromUUID` are synchronous file reads that were blocking the render thread on every drag-drop. Both now run on a `ThreadPool` worker thread using a dedicated per-drop `ArenaAllocator` (4× file size + 8 MB) that outlives the lambda. The main-thread callback via `MainThreadScheduler::Post` handles only the lightweight `IngestMesh` + actor create.

## Result

Stable 121 fps through mesh drag-drop with textures. Multiple meshes rendered correctly.

Viewport with dragged meshes

## Test plan

- Drag a cooked `.zemesh` asset onto the viewport — FPS stays stable, mesh appears with correct textures
- Drag multiple meshes in sequence — no frame drops, all visible
- Verify on Windows: no Intel timeline semaphore corruption (`UINT64_MAX`) with `vkQueueSubmit2`


JeanPhilippeKernel:
Closing — duplicate of #773 which targets develop correctly.

Issue #714, opened 2026-09-01
feat(scene): stable component schema registry


JeanPhilippeKernel:
## Summary

Evolve the existing runtime `ComponentSerializerRegistry` into the stable,
versioned component-schema registry required by the scene document contract.
The current registry dispatches C-style YAML/binary callbacks by
`ComponentTypeID`; that ID is process-local and registration-order dependent, so
it must never appear as a durable file-format identifier.

## Scope

- Give every serializable component a stable schema key and schema version.
- Associate the runtime `ComponentTypeID` only after reading the stable key.
- Register YAML and binary codecs, compatibility/migration capability, authored
  versus runtime/editor-only classification, and UUID-reference
  enumerate/patch hooks.
- Define stable field keys for inspector/history codec values without treating
  reflection offsets or raw bytes as a serialization schema.
- Make initialization and built-in registration explicit during engine startup;
  static registration order is not accepted.
- Provide deterministic iteration by canonical schema key for save/cook output.
- Prefer `cstring` for new engine-facing non-ABI string parameters.

## Non-goals

- Do not serialize a raw `ComponentTypeID`, memory layout, pointer, or field
  offset.
- Do not implement the built-in component codecs here; #717 owns those.
- Do not define scene-document YAML, loading, or cooked-file layout here.

## Acceptance criteria

- The registry is initialized before any serializer is offered.
- A saved document contains only stable component keys and versions.
- Duplicate/invalid schema registration has a bounded diagnostic and cannot
  produce nondeterministic output.
- The registry can enumerate/patch UUID references without each loader inventing
  component-specific reflection rules.
- Unit tests cover explicit registration, stable ordering, lookup, and rejected
  invalid registrations.

## References

`ZEngine/docs/future-plan/scene-serialization.md` sections 2, 4, 8, and 11.

## Follow-on work

- #715 — versioned YAML SceneDocument and canonical save
- #717 — built-in component codecs and reference hooks
- #830 — unknown-data and migration policy

PR #837, opened 2026-09-20
feat(scene): stable component schema registry

jnyfah:


ZEngine/ZEngine/ECS/SceneComponentSchemaRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+            ZENGINE_CORE_ERROR("SceneComponentSchemaRegistry::Register rejected: {}", message)
+        }
+    } // namespace
+
+    bool SceneComponentSchemaRegistry::IsInitialized() const
+    {
+        return m_arena != nullptr;
+    }
+
+    void SceneComponentSchemaRegistry::Initialize(Core::Memory::ArenaAllocator* arena, uint32_t capacity)
```
JeanPhilippeKernel: Blocking: #714 requires this registry to be initialized explicitly during engine startup before any serializer is offered. This PR adds no EngineContext owner or Engine::Initialize call, and YAMLSceneSerializer still uses the legacy ComponentSerializerRegistry keyed by process-local ComponentTypeID. Please wire the new registry into startup and the serializer boundary before merging.

ZEngine/ZEngine/ECS/SceneComponentSchemaRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+        ZENGINE_VALIDATE_ASSERT(arena != nullptr, "SceneComponentSchemaRegistry::Initialize: arena must not be null")
+        ZENGINE_VALIDATE_ASSERT(capacity > 0, "SceneComponentSchemaRegistry::Initialize: capacity must be > 0")
+
+        if (m_arena)
+        {
+            return;
+        }
+
+        m_arena = arena;
+        m_schemas.init(arena, capacity);
```
JeanPhilippeKernel: Please do not mark the registry initialized until both arrays have allocated successfully. Array::reserve records the requested capacity even when the arena allocation returns nullptr, so an exhausted arena leaves IsInitialized() true and the first Register() writes through a null m_schemas buffer. Add an explicit allocation check/assertion (and coverage with a deliberately undersized arena).

ZEngine/ZEngine/ECS/SceneComponentSchemaRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+        schema.Key         = CopyKey(desc.Key);
+        schema.Version     = desc.Version;
+        schema.RuntimeType = desc.RuntimeType;
+        schema.FieldCount  = desc.FieldCount;
+        schema.Codecs      = desc.Codecs;
+        schema.References  = desc.References;
+
+        if (desc.FieldCount > 0)
+        {
+            auto* fields = ZPushArray(m_arena, SceneFieldSchema, desc.FieldCount);
```
JeanPhilippeKernel: Please handle allocation failure before iterating. ZPushArray can return nullptr, after which fields[i] dereferences it; CopyKey can also fail and currently leaves a schema with null wire keys. Register must return false with a bounded diagnostic and publish no partial schema when either allocation fails.

ZEngine/ZEngine/ECS/SceneComponentSchemaRegistry.cpp
```diff
@@ -0,0 +1,229 @@
...
+            for (uint32_t i = 0; i < desc.FieldCount; ++i)
+            {
+                fields[i].Key   = CopyKey(desc.Fields[i].Key);
+                fields[i].Class = desc.Fields[i].Class;
```
JeanPhilippeKernel: Please validate that Class is one of Authored, RuntimeDerived, EditorOnly, or Forbidden before storing it. enum class values can still be constructed from arbitrary serialized/plugin input; accepting an unknown value creates a schema the save/load policy cannot interpret. Add a rejection diagnostic and a test.

JeanPhilippeKernel (changes requested):
Requesting changes for the four inline findings: engine/serializer lifecycle integration, initialization allocation handling, registration allocation handling, and SceneFieldClass validation. These are required for #714 acceptance.

